PR #5, opened 2017-12-12
Exclude already finished events

mic4ael:


outlook/indico_outlook/util.py
```diff
@@ -27,7 +27,8 @@ def check_config():
 def is_event_excluded(event):
     """Check if an event is excluded from the calendar"""
     from indico_outlook.plugin import OutlookPlugin
-    return event.duration > OutlookPlugin.settings.get('max_event_duration')
+    has_event_ended = event.end_dt is not None and event.end_dt <= now_utc()
```
ThiefMaster: In which cases would the `end_dt` be None? The column is not nullable.

mic4ael: indeed.. thanks!

PR #10, opened 2018-03-01
Avoid new transactions on a complete registration

mic4ael:


mvidalgarcia:
`Payment/CERN: ` as commit message prefix?

mic4ael:
Thanks! I always forget about that part.

PR #11, opened 2018-03-12
CERNAccess: require at least one reg. form

pferreir:
Requests with no selected registration forms are confusing.
Also added a message box pointing to the docs once a request is sent/accepted.


ThiefMaster:
I'd put it `1.0-maintenance` - at least in case you want to put it in prod before 2.1?

PR #13, opened 2018-04-03
Send correct query to the Search service

mic4ael:


search_cern/indico_search_cern/engine.py
```diff
@@ -53,7 +55,14 @@ def _make_query(self):
             if field in FIELD_MAP:
                 query += self._make_field_query(phrase, FIELD_MAP[field])
             else:
-                query.append(phrase)
+                pattern = re.compile(r'^(-?)({}):'.format('|'.join(FIELD_MAP.iterkeys())), re.IGNORECASE)
```
ThiefMaster: no need for `.iterkeys()`. iterating a dict already gives you its keys.

ThiefMaster:
bump version to 1.0.1

PR #14, opened 2018-04-17
CERNAccess: Truncate event title for group name

meluru:


cern_access/indico_cern_access/util.py
```diff
@@ -100,13 +100,14 @@ def build_access_request_data(registration, event, generate_code):
     from indico_cern_access.plugin import CERNAccessPlugin
     start_dt, end_dt = get_access_dates(get_last_request(event))
     tz = timezone('Europe/Zurich')
+    title = (event.title[:97] + '...') if len(event.title > 100) else event.title
```
ThiefMaster: `event.title > 100` will fail ;)

You could just use the util from `jinja2.filters`:

    title = do_truncate(event.title, 100, leeway=0)

PR #34, opened 2019-02-15
CERNAccess: Fix bug on user update

pferreir:
Participants were being set as active erroneously.

### DIFF
--- a/cern_access/indico_cern_access/plugin.py
+++ b/cern_access/indico_cern_access/plugin.py
@@ -242,12 +242,15 @@ class CERNAccessPlugin(IndicoPlugin):
         """
         if (registration_form.cern_access_request and registration_form.cern_access_request.is_active and
                 not self._is_past_event(registration_form.event)):
-            registrations = get_requested_registrations(registration_form.event, regform=registration_form)
-            if registrations:
-                deleted = send_adams_delete_request(registrations)
-                if deleted:
-                    withdraw_access_requests(registrations)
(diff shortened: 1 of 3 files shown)


cern_access/indico_cern_access/plugin.py
```diff
@@ -242,12 +242,13 @@ def _registration_form_deleted(self, registration_form, **kwargs):
...
-            registrations = get_requested_registrations(registration_form.event, regform=registration_form)
-            if registrations:
-                deleted = send_adams_delete_request(registrations)
-                if deleted:
-                    withdraw_access_requests(registrations)
-                    notify_access_withdrawn(registrations)
+            requests = get_requested_registrations(registration_form.event, regform=registration_form)
+            if requests:
+                active_requests = [r for r in requests if r.state == CERNAccessRequestState.active]
+                # XXX: what do do here? should we check the return value?
```
ThiefMaster: log the error and maybe even raise one to prevent the change from being committed? that way the user will likely try again

pferreir: Hm... What kind of error should we raise? NoReport?

pferreir: No, there should be a report option, that's clear.

ThiefMaster: `RuntimeError` maybe?

ThiefMaster: or just Exception?

PR #37, opened 2019-02-20
Startup assistance

mic4ael:


startup_assistance/indico_startup_assistance/models/startup_assistance_requests.py
```diff
@@ -0,0 +1,38 @@
...
+        backref=db.backref(
+            'startup_assistance_request',
+            uselist=False,
+            lazy=False
+        )
+    )
+
+    @return_ascii
+    def __repr__(self):
+        return '<StartupAssistanceRequest({})'.format(self.reservation_id)
```
ThiefMaster: `format_repr`

startup_assistance/indico_startup_assistance/models/startup_assistance_requests.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    registration = db.relationship(
+        'Reservation',
+        uselist=False,
+        lazy=True,
+        backref=db.backref(
+            'startup_assistance_request',
+            uselist=False,
+            lazy=False
```
ThiefMaster: `lazy=True`

startup_assistance/indico_startup_assistance/models/startup_assistance_requests.py
```diff
@@ -0,0 +1,38 @@
...
+
+    reservation_id = db.Column(
+        db.ForeignKey('roombooking.reservations.id'),
+        primary_key=True
+    )
+
+    registration = db.relationship(
+        'Reservation',
+        uselist=False,
+        lazy=True,
```
ThiefMaster: you probably want `lazy=False` here  so you get the registrations when fetching the requests for the list of active assistance requests

startup_assistance/indico_startup_assistance/operations.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def create_startup_assistance_request(reservation):
+    startup_request = StartupAssistanceRequest()
+    startup_request.reservation = reservation
+    db.session.flush()
+
+
+def cancel_startup_assistance_request(reservation):
+    if reservation.startup_assistance_request is not None:
```
ThiefMaster: `reservation.startup_assistance_request = None` probably does the job

startup_assistance/indico_startup_assistance/operations.py
```diff
@@ -0,0 +1,24 @@
...
+
+from __future__ import unicode_literals
+
+from indico.core.db import db
+
+from indico_startup_assistance.models.startup_assistance_requests import StartupAssistanceRequest
+
+
+def create_startup_assistance_request(reservation):
+    startup_request = StartupAssistanceRequest()
```
ThiefMaster: reservation.whatever = StartupAssistanceRequest()

and i think this whole thing is now so small that it can just go directly in wherever you call these functiosn here

startup_assistance/indico_startup_assistance/forms.py
```diff
@@ -0,0 +1,17 @@
...
+
+from __future__ import unicode_literals
+
+from indico.web.forms.base import IndicoForm
+from indico.web.forms.fields import EmailListField
+
+from indico_startup_assistance import _
+
+
+class StartupAssistanceForm(IndicoForm):
```
ThiefMaster: i'd keep this in plugin.py - this is small and most plugins don't have a separate forms.py for it

startup_assistance/indico_startup_assistance/plugin.py
```diff
@@ -0,0 +1,61 @@
...
+    """Startup assistance request
+
+    Provides a service request where users can ask
+    for startup assistance with their booking.
+    """
+
+    configurable = True
+    settings_form = StartupAssistanceForm
+    default_settings = {
+        'startup_assistance_recipients': set(),
```
ThiefMaster: `[]` since the field will give you a list as well

startup_assistance/indico_startup_assistance/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# The CERN Indico plugins are free software; you can redistribute
+# them and/or modify them under the terms of the MIT License; see
+# the LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from indico.util.i18n import make_bound_gettext
+
+
+_ = make_bound_gettext('startup_assistance')
```
ThiefMaster: Wouldn't something like `room_assistance` be a better name for the plugin?

startup_assistance/indico_startup_assistance/client/js/index.js
```diff
@@ -0,0 +1,13 @@
...
+ * The CERN Indico plugins are free software; you can redistribute
+ * them and/or modify them under the terms of the MIT License; see
+ * the LICENSE file for more details.
+ */
+
+import setup from 'indico/modules/rb_new/setup';
+
+import overrides from './overrides';
+
+setup({...overrides});
```
ThiefMaster: doesn't this mean we call `setup()` twice now? in burotel it's fine because we *replace* the whole room booking, but here we just want to extend it and certainly not duplicate the whole RB stuff in the plugin (otherwise the plugin needs to be rebuilt whenever something in RB changes)

startup_assistance/indico_startup_assistance/controllers.py
```diff
@@ -0,0 +1,22 @@
...
+from indico.web.rh import RHProtected
+
+from indico_startup_assistance.views import WPStartupAssistance
+
+
+class RHRequestList(RHProtected):
+    """Provides a list of videoconference assistance requests"""
+
+    def _check_access(self):
+        RHProtected._check_access(self)
```
ThiefMaster: remove this unless you add something else here at some point

startup_assistance/indico_startup_assistance/controllers.py
```diff
@@ -0,0 +1,22 @@
+# This file is part of the CERN Indico plugins.
+# Copyright (C) 2014 - 2018 CERN
```
ThiefMaster: let's use 2019 in newly added files

startup_assistance/indico_startup_assistance/definition.py
```diff
@@ -0,0 +1,17 @@
...
+# the LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from indico.modules.events.requests import RequestDefinitionBase
+
+from indico_startup_assistance import _
+
+
+class StartupAssistanceRequest(RequestDefinitionBase):
```
ThiefMaster: obsolete

startup_assistance/indico_startup_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,68 @@
...
+        AND end_dt > NOW()::TIMESTAMP
+        AND state IN (1, 2)
+    ''')
+    for row in res:
+        conn.execute('''
+            INSERT INTO plugin_startup_assistance.startup_assistance_requests
+            VALUES (%s)
+        ''', (row.id,))
+
+    op.drop_column('reservations', 'needs_assistance', schema='roombooking')
```
ThiefMaster: :warning: :bomb: this must be in the core

startup_assistance/indico_startup_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,68 @@
...
+    op.create_table(
+        'startup_assistance_requests',
+        sa.Column('reservation_id', sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(['reservation_id'], ['roombooking.reservations.id']),
+        sa.PrimaryKeyConstraint('reservation_id'),
+        schema='plugin_startup_assistance'
+    )
+
+    conn = op.get_bind()
+    res = conn.execute('''
```
ThiefMaster: i think you can turn this into one query that does the insert for each matching row. that way you can even use `op.execute()` and don't need to force online mode

startup_assistance/indico_startup_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,68 @@
...
+    if context.is_offline_mode():
+        raise Exception('This downgrade is only possible in online mode')
+
+    op.add_column('reservations',
+                  sa.Column('needs_assistance', sa.Boolean(), nullable=False, server_default='false'),
+                  schema='roombooking')
+
+    conn = op.get_bind()
+    res = conn.execute('SELECT reservation_id FROM plugin_startup_assistance.startup_assistance_requests')
+    for row in res:
```
ThiefMaster: just assume the column in reservation exists and then do something like `UPDATE reservations SET ... = true WHERE EXISTS (...)` in a single query

startup_assistance/indico_startup_assistance/plugin.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def get_blueprints(self):
+        return blueprint
+
+    def _extend_services_menu(self, reservation, **kwargs):
+        return TopMenuItem('services-cern-startup-assistance', _('Startup assistance'),
+                           url_for_plugin('startup_assistance.request_list'), section='services')
+
+    def _create_request_if_necessary(self, reservation, **kwargs):
+        extra_fields = kwargs.get('extra_fields', {})
```
ThiefMaster: add `extra_fields` as a normal arg, `**kwargs` in a signal handler is only for forward compatibility if we add additional arguments

startup_assistance/indico_startup_assistance/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+from indico.core.plugins import get_plugin_template_module
+from indico.modules.rb.models.reservation_occurrences import ReservationOccurrence
+from indico.modules.rb.models.reservations import Reservation
+
+from indico_startup_assistance.plugin import StartupAssistanceRequestPlugin
+
+
+def _send_email(recipients, template):
+    email = make_email(from_address='noreply-indico-team@cern.ch', to_list=recipients, template=template, html=True)
+    send_email(email, module='Indico Startup Assistance CERN')
```
ThiefMaster: no need to set a module, it's only used when sending an email within an event (it's used for the event log)

startup_assistance/setup.py
```diff
@@ -0,0 +1,35 @@
...
+# the LICENSE file for more details.
+
+from __future__ import unicode_literals
+
+from setuptools import find_packages, setup
+
+
+setup(
+    name='indico-plugin-startup-assistance',
+    version='1.0',
```
ThiefMaster: `3.0-dev` for now

startup_assistance/indico_startup_assistance/client/js/index.js
mic4ael: Oh, that is true! I think I shouldn't do that

startup_assistance/indico_startup_assistance/operations.py
mic4ael: Yeah, yeah. Just like I mentioned. Initially I did it in a separate method because I didn't know how big the function will be. I will definitely move it back to `plugin.py` if necessary.

startup_assistance/indico_startup_assistance/__init__.py
mic4ael: Hm, could be. I just sticked to how it was called before.

startup_assistance/indico_startup_assistance/controllers.py
mic4ael: This is a draft, eventually I will either add something there or remove it altogether.

startup_assistance/indico_startup_assistance/controllers.py
mic4ael: Ok, I thought that we said we should keep it 2018 and just before release we will change all places in one go.

startup_assistance/indico_startup_assistance/definition.py
mic4ael: Yeah, yeah ;)

startup_assistance/indico_startup_assistance/controllers.py
ThiefMaster: yep, just added the comment to not forget it later in case it stays like this

startup_assistance/indico_startup_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
mic4ael: Yes, I will move it as discussed.

startup_assistance/indico_startup_assistance/controllers.py
ThiefMaster: yep we will update all the existing files in one go, but nothing wrong with updating the file template in your editor to use 2019 for new files :)

room_assistance/indico_room_assistance/plugin.py
```diff
@@ -0,0 +1,104 @@
...
+        super(RoomAssistancePlugin, self).init()
+        self.template_hook('event-actions', self._room_assistance_action)
+        self.connect(signals.menu.items, self._extend_services_menu, sender='top-menu')
+        self.connect(signals.plugin.get_event_request_definitions, self._get_room_assistance_request)
+
+    def get_blueprints(self):
+        return blueprint
+
+    def _room_assistance_action(self, event, **kwargs):
+        has_room_assistance_request = event.requests.filter_by(type='room-assistance').count() > 0
```
ThiefMaster: `.has_rows()`

room_assistance/indico_room_assistance/tasks.py
```diff
@@ -0,0 +1,39 @@
...
+from indico_room_assistance.plugin import RoomAssistancePlugin
+
+
+@celery.periodic_task(run_every=crontab(minute='0', hour='6'), plugin='room_assistance')
+def room_assistance_emails():
+    requests = (Request.query
+                .filter(Request.type == 'room-assistance',
+                        Request.state == RequestState.accepted,
+                        Event.own_room_id.isnot(None),
+                        db.cast(Request.data.op('->>')('start_dt'), db.Date) == db.cast(now_utc(), db.Date))
```
ThiefMaster: I think you can use `Request.data['start_dt'].astext` here instead of calling the operator manually.

room_assistance/indico_room_assistance/templates/emails/modified_request_email_to_assistance.html
```diff
@@ -0,0 +1,20 @@
...
+
+{% block subject_text -%}
+    Room assistance request has been modified
+{%- endblock %}
+
+{% block body -%}
+    <p>Startup assistance for the event {{ event.title }} has been modified.</p>
+
+    <div>
+        {% if changes['start_dt'] -%}
```
ThiefMaster: `changes.start_dt` etc.

room_assistance/indico_room_assistance/templates/emails/modified_request_email_to_assistance.html
```diff
@@ -0,0 +1,20 @@
...
+            <strong>When:</strong> {{ changes['start_dt']['old']|format_datetime }} → {{ changes['start_dt']['new']|format_datetime }}<br>
```
ThiefMaster: wrap after `</strong>`

room_assistance/indico_room_assistance/templates/event_request_details.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'events/requests/event_request_details.html' %}
+{% from 'message_box.html' import message_box %}
+
+{% block content %}
+    {% if not event_has_room_attached %}
+         {% call message_box('error') %}
+            {% trans -%}
+                This event doesn't have location specified thus you can't request room assistance
```
ThiefMaster: a location

room_assistance/indico_room_assistance/templates/event_request_details.html
```diff
@@ -0,0 +1,20 @@
...
+            {%- endtrans %}
+        {% endcall %}
+    {% elif not room_allows_assistance %}
+        {% call message_box('error') %}
+            {% trans room_name=event.room.full_name -%}
+                Room assistance can't be requested for the event location (<strong>{{ room_name }}</strong>)
```
ThiefMaster: s/can't/cannot/?

room_assistance/indico_room_assistance/templates/event_request_details.html
ThiefMaster: s/doesn't/does not/

room_assistance/indico_room_assistance/templates/request_list.html
```diff
@@ -0,0 +1,59 @@
...
+            <td>{{ request.event.room.full_name }}</td>
+            <td>
+                <a href="{{ request.event.url }}">
+                    {{ request.event.title }}
+                </a>
+            </td>
+            <td>
+                {{ parse_dt(request.data['start_dt']) | format_datetime }}
+            </td>
+            <td>{{ request.data['reason'] }}</td>
```
ThiefMaster: `request.data.reason` (same for the other places where bracket syntax is used)

room_assistance/indico_room_assistance/templates/request_list.html
```diff
@@ -0,0 +1,59 @@
...
+        </tr>
+    </tbody>
+{% endmacro %}
+
+
+{% block content %}
+    {{ super() }}
```
ThiefMaster: don't override a block if you just call super ;)

room_assistance/indico_room_assistance/tasks.py
mic4ael: Mhm, that I didn't know. I will check!

mic4ael: Nice, that works ;)

room_assistance/indico_room_assistance/templates/request_list.html
mic4ael: WTH, it wasn't intentional.. I just forgot to remove it.

room_assistance/indico_room_assistance/plugin.py
```diff
@@ -0,0 +1,94 @@
...
+    return sorted(object_list, key=lambda r: natural_sort_key(r[1].full_name))
+
+
+class RoomAssistanceForm(IndicoForm):
+    _fieldsets = [
+        ('Startup assistance emails', ['room_assistance_recipients', 'rooms_with_assistance',
+                                       'room_assistance_support']),
+    ]
+
+    room_assistance_recipients = EmailListField(_('Recipients'))
```
mvidalgarcia: I think it would make sense to add a description of this field as there must be one email address per line, no?
Something like "Notifications about room assistance requests are sent to these email addresses (one per line)"?

ThiefMaster: Seems pretty obvious to me. Anyway, only we configure this plugin.

mvidalgarcia: At first I would have added a list of email addresses separated by commas.

mic4ael: Yeah, but if that is configured by us, then it is not mandatory. However, it deosn't harm to add it ;) I will add a short description.

room_assistance/indico_room_assistance/templates/event_request_details.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'events/requests/event_request_details.html' %}
+{% from 'message_box.html' import message_box %}
+
+{% block content %}
+    {% if not event_has_room_attached %}
+         {% call message_box('error') %}
+            {% trans -%}
+                This event does not have a location specified thus you cannot request room assistance
```
mvidalgarcia: Link to event settings?

mic4ael: How would that help if you have the link on the left side? :P

mvidalgarcia: I had it in the `vc_assistance` plugin, I think it was suggested by Thomas. I thought this would be the same.

mic4ael: Hm, to me it looks like an unnecessary repetition of something that is already clearly available.

mvidalgarcia: Ok... nevermind.

mic4ael: Maybe I should add it anyway for the sake of consistency. Thanks anyway for having a look at this!

room_assistance/indico_room_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,90 @@
...
+    table_data = [['ID', 'Room ID', 'Link ID', 'Start date', 'End date', 'Booked for']]
+    for resv in invalid_reservations:
+        table_data.append(list(resv))
+
+    if len(table_data) > 1:
+        print(AsciiTable(table_data,
+                         'Reservations without a link to an event that need assistance in the future').table)
+
+    conn.execute('''
+        UPDATE settings SET module = 'plugin_room_assistance', name = 'room_assistance_recipients'
```
ThiefMaster: did you test this? the table names here are missing the schema so the queries fail, so if this code actually works for you something is screwed up with your postgres search_path

room_assistance/indico_room_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,90 @@
...
+    invalid_reservations = conn.execute('''
+        SELECT rv.id, rv.room_id, rvl.id, rv.start_dt, rv.end_dt, rv.booked_for_name
+        FROM roombooking.reservations rv
+        LEFT JOIN roombooking.reservation_links rvl ON rv.link_id = rvl.id
+        WHERE rv.needs_assistance
+        AND (rv.link_id IS NULL OR rvl.link_type != 2)
+        AND rv.start_dt > NOW()
+    ''')
+
+    table_data = [['ID', 'Room ID', 'Link ID', 'Start date', 'End date', 'Booked for']]
```
ThiefMaster: Link ID is always None, because your never have a reservation link when migrating the data

ThiefMaster: What about printing the room name instead of ID? If we sent the current list to the assistance guys they'd probably hate us

room_assistance/indico_room_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,90 @@
...
+        raise Exception('This upgrade is only possible in online mode')
+
+    conn = op.get_bind()
+    conn.execute('''
+        INSERT INTO events.requests (event_id, type, state, data, created_by_id, created_dt)
+        SELECT
+            rvl.linked_event_id,
+            'room-assistance',
+            1,
+            '{}',
```
ThiefMaster: This goes kaboom when trying to view the request details page in the given event. We shouldn't migrate this with empty data...

room_assistance/indico_room_assistance/notifications.py
```diff
@@ -0,0 +1,52 @@
...
+    event = request.event
+    return '{prefix}[{room}] {subject}'.format(
+        prefix='[Support Request]',
+        room=event.room.full_name,
+        subject=subject
+    ).strip()
+
+
+def send_email_to_assistance(request, template_name, **template_params):
+    to_list = rb_settings.get('assistance_emails')
```
ThiefMaster: This should be moved to the plugin settings.

mic4ael: Good catch! I should just remove it and use the setting from the plugin, because I already have a setting for it.

room_assistance/indico_room_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
mic4ael: isn't the `indico` schema the default one?

ThiefMaster: Now I remember, the search path is `"$user", public` by default. So if your user is called indico it works for you. That's why your DB user in development should not be called indico ;)

mic4ael: Oh. I should have checked that..

room_assistance/indico_room_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
mic4ael: Are you sure LINK ID is always None? We have a migration that migrates that data in the core.

ThiefMaster: `AND (rv.link_id IS NULL OR rvl.link_type != 2)` - I guess 2 is event, and we only migrate event links since any other links did not exist before

room_assistance/indico_room_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,100 @@
...
+        FROM roombooking.reservations rv
+        JOIN roombooking.reservation_links rvl ON rv.link_id = rvl.id AND rvl.link_type = 2
+        WHERE rv.needs_assistance
+    ''')
+
+    invalid_reservations = conn.execute('''
+        SELECT
+            rv.id,
+            rv.room_id,
+            format('%%s/%%s-%%s', r.building, r.floor, r.number),
```
mvidalgarcia: Why double `%`?

mic4ael: To escape `%` so that it is not interpreted as a part of a formatted string.

ThiefMaster: In Python? Unless you use `%` after the string, you do not need to escape it..

ThiefMaster: Since I first thought "why, there's no `%` formatting applied on the string": `%s` is also used by the Postgres DBAPI for params.

mvidalgarcia: It's strange because I don't need to escape it in my script :thinking:

ThiefMaster: I guess you aren't passing it to the DBAPI since you are using the models?

mic4ael:
@mvidalgarcia thanks for having a look ;)

room_assistance/indico_room_assistance/controllers.py
```diff
@@ -0,0 +1,63 @@
...
+from indico_room_assistance.util import is_room_assistance_support
+from indico_room_assistance.views import WPRoomAssistance
+
+
+def _find_requests(from_dt=None, to_dt=None):
+    query = (Request.query
+             .options(joinedload(Request.event))
+             .filter(Request.type == 'room-assistance',
+                     Request.state == RequestState.accepted,
+                     db.cast(Request.data, postgresql.JSONB).op('?')('start_dt')))
```
ThiefMaster: `db.cast(Request.data, postgresql.JSONB).has_key('start_dt')` should work as well.

BTW, what about converting the column to JSONB in the core?

room_assistance/indico_room_assistance/forms.py
```diff
@@ -0,0 +1,62 @@
+# This file is part of the CERN Indico plugins.
+# Copyright (C) 2014 - 2018 CERN
```
ThiefMaster: 2019

room_assistance/indico_room_assistance/forms.py
```diff
@@ -0,0 +1,62 @@
...
+from indico.web.forms.base import IndicoForm, generated_data
+from indico.web.forms.fields import IndicoDateField, IndicoDateTimeField
+from indico.web.forms.validators import DateTimeRange, Exclusive
+
+from indico_room_assistance import _
+
+
+class RoomAssistanceRequestForm(RequestFormBase):
+    start_dt = IndicoDateTimeField(_('When'), [DataRequired(),
+                                               DateTimeRange(earliest=datetime.combine(now_utc().date(), time()))],
```
ThiefMaster: why not just `datetime.now()`? or even better, get the current localtime in the instance's DEFAULT_TIMEZONE. however, since it's a CERN plugin and the server will always be in the same timezone, `datetime.now()` should be fine.

room_assistance/indico_room_assistance/forms.py
```diff
@@ -0,0 +1,62 @@
...
+                                   description=_('When do you need the assistance?'))
+    reason = TextAreaField(_('Reason'), [DataRequired()], description=_('Why are you requesting assistance?'))
+
+    def validate_start_dt(self, field):
+        localized_time = field.data.astimezone(session.tzinfo).time()
+        is_in_working_hours = any(period[0] <= localized_time <= period[1] for period in Location.working_time_periods)
```
ThiefMaster: needs to be updated. feel free to copy the working time periods into this plugin. having "working time periods" hardcoded in the core isn't amazing anyway, so if we ever make them more dynamic for the occupancy calculation there we don't have to worry about updating the plugin as well.

room_assistance/indico_room_assistance/migrations/20190220_1138_b160812c186b_add_startup_assistance_requests_table.py
```diff
@@ -0,0 +1,100 @@
+"""Add room assistance requests table
```
ThiefMaster: iirc. we wanted to move this to a standalone script like for vc assistance

ThiefMaster: while using the models in such a script would be cleaner, feel free to keep the SQL since that way it's minimal work to move it to the standalone script

room_assistance/indico_room_assistance/notifications.py
```diff
@@ -0,0 +1,44 @@
...
+from indico.core.notifications import make_email, send_email
+from indico.core.plugins import get_plugin_template_module
+from indico.modules.events.requests.models.requests import RequestState
+
+
+def send_email_to_assistance(request, template_name, **template_params):
+    from indico_room_assistance.plugin import RoomAssistancePlugin
+
+    to_list = RoomAssistancePlugin.settings.get('room_assistance_recipients')
+    if to_list:
```
ThiefMaster:
```
if not to_list:
    return
```

room_assistance/indico_room_assistance/notifications.py
```diff
@@ -0,0 +1,44 @@
...
+                                              **template_params)
+        send_email(make_email(from_address=config.NO_REPLY_EMAIL, to_list=to_list, template=template, html=True))
+
+
+def notify_about_new_request(request):
+    send_email_to_assistance(request, template_name='creation_email_to_assistance')
+
+
+def notify_about_request_modification(request, changes):
+    if not request.state == RequestState.accepted:
```
ThiefMaster: `not ==`? If Python just had an operator that checks for inequality... :trollface:

room_assistance/indico_room_assistance/notifications.py
```diff
@@ -0,0 +1,44 @@
...
+        return
+    send_email_to_assistance(request, template_name='modified_request_email_to_assistance', changes=changes)
+
+
+def notify_about_withdrawn_request(request):
+    if not request.state == RequestState.withdrawn:
```
ThiefMaster: 🙈

room_assistance/indico_room_assistance/notifications.py
mic4ael: ohh shit what did I do here

room_assistance/indico_room_assistance/notifications.py
mic4ael: OMG..

mic4ael: That is what happens when you write more JS than Python..

room_assistance/indico_room_assistance/controllers.py
mic4ael: Hm, I thought I changed all the occurrences of that.

room_assistance/indico_room_assistance/forms.py
mic4ael: I would pass `now_utc()` since in the `DateTimeRange` the value passed to `earliest` is naively localized to `UTC` anyway. So passing `datetime.now()` would create a small inconsistency.

mic4ael:
Migration script: https://gist.github.com/mic4ael/4983e37cc65b647a60d4a548c8cec0f1

PR #39, opened 2019-02-21
CERN Access tests

pferreir:


ThiefMaster:
tests shouldn't have a `__init__.py`

pferreir:
> tests shouldn't have a `__init__.py` and be outside the package

I need an `__init__.py` to be able to import stuff.

cern_access/indico_cern_access/tests/operations_test.py
```diff
@@ -0,0 +1,129 @@
...
+def dummy_access_request(dummy_regform):
+    """Create a registration and corresponding request."""
+    form = make_registration_form(dummy_regform)(csrf_enabled=False)
+    form.validate()
+    return create_registration(dummy_regform, form.data).cern_access_request
+
+
+def setup_fixtures(func):
+    """Set up fixtures (utlity decorator)."""
+    pytest.mark.usefixtures('smtp', 'mock_access_request', 'dummy_access_request')(func)
```
ThiefMaster: shouldn't there be `func = ...` even in case the func isn't actually replaced with something new?

cern_access/indico_cern_access/tests/request_test.py
```diff
@@ -0,0 +1,93 @@
...
+    assert form.validate()
+
+    registration = create_registration(dummy_regform, form.data)
+    assert registration.cern_access_request is None
+
+
+@pytest.mark.parametrize('mock_access_request',
+                         [{
+                             'mode': RegformDataMode.during_registration_required
+                          }],
```
ThiefMaster: too much indentation (probably easier to just have the whole `[{...}]` part in one line here)

pferreir: Why is it too much indentation?

cern_access/indico_cern_access/plugin.py
```diff
@@ -227,7 +227,9 @@ def _registration_form_validated(self, form, **kwargs):
 
         required = req.data.get('regform_data_mode') == RegformDataMode.during_registration_required
         form_cls = AccessIdentityDataForm if required else RegistrationFormPersonalDataForm
-        g.personal_data_form = form = form_cls()
+        # we pass 'csrf_enabled' on to allow unit tests to set it to False
+        g.personal_data_form = form = form_cls(csrf_enabled=form.meta.csrf)
```
ThiefMaster: i think it could always be disabled here since 1) we have a global csrf check and 2) the regform probably has a csrf check as well

pferreir: Does it hurt to leave it like this?

cern_access/indico_cern_access/tests/request_test.py
ThiefMaster:
```python
@pytest.mark.parametrize('mock_access_request',
                         [{
                             'mode': RegformDataMode.during_registration_required,
                             'personal_data': PERSONAL_DATA
                          }],
                         indirect=True)
def foo():
    pass
```

vs

```python
@pytest.mark.parametrize('mock_access_request',
                         [{
                             'mode': RegformDataMode.during_registration_required,
                             'personal_data': PERSONAL_DATA
                         }],
                         indirect=True)
def foo():
    pass
```

cern_access/indico_cern_access/tests/operations_test.py
pferreir: I guess it depends on what the decorator does. This one doesn't wrap the function, from what I see. But yeah, it makes more sense.

cern_access/indico_cern_access/plugin.py
ThiefMaster: nope

cern_access/indico_cern_access/tests/operations_test.py
ThiefMaster: yeah, just to avoid problems if pytest ever decides to replace the function with something else

ThiefMaster:
Also update `.travis.yml` to actually run these tests. Might want to add a pytest.ini as well so `run-tests.sh` runs them too.

cern_access/indico_cern_access/tests/request_test.py
pferreir: Ah, right. I hadn't noticed that. I wonder why flake8 doesn't complain...

pferreir:
Should work OK against my [PR for the core](https://github.com/indico/indico/pull/3794).

PR #45, opened 2019-03-18
Update FoundationSync

panagiotappl:
closes indico/indico#3818

### DIFF
--- a/foundationsync/indico_foundationsync.py
+++ b/foundationsync/indico_foundationsync.py
@@ -81,9 +81,6 @@ class FoundationSync(object):
             self._logger.exception("Problem connecting to DB")
             raise
 
-    def _get_room_attrs(self, raw_data):
-        return {'manager-group': raw_data.get('EMAIL_LIST')}
-
     def _html_to_markdown(self, s):
         s = re.sub(r'<font color=[^> ]+>(.+?)</font>', r'<strong>\1</strong>', s)
         return HTML2Text(bodywidth=0).handle(s).strip()
@@ -133,17 +130,12 @@ class FoundationSync(object):
     def _prepare_row(self, row, cursor):
         return dict(zip([d[0] for d in cursor.description], row))
 
-    def _update_room(self, room, room_data, room_attrs):
+    def _update_room(self, room, room_data):
         room.is_active = True
         room.is_reservable = True
         for k, v in room_data.iteritems():
             if getattr(room, k) != v:
                 setattr(room, k, v)
-        for attribute, value in room_attrs.iteritems():
-            if value:
-                value = value.strip()
-            if room.get_attribute_value(attribute) != value:
-                room.set_attribute_value(attribute, value)
         db.session.flush()
 
     def fetch_buildings_coordinates(self, connection):
@@ -211,7 +203,7 @@ class FoundationSync(object):
 
             try:
                 room_data, email_warning = self._parse_room_data(data, coordinates, room_id)
-                room_attrs = self._get_room_attrs(data)
+                manager_group = data.get('EMAIL_LIST')
                 self._logger.debug("Fetched data for room with id='%s'", room_id)
             except SkipRoom as e:
                 counter['skipped'] += 1
@@ -240,11 +232,10 @@ class FoundationSync(object):
                 counter['updated'] += 1
 
             # Update room data
-            self._update_room(room, room_data, room_attrs)
+            self._update_room(room, room_data)
             new_managers = set()
             if room_data.get('owner') is not None:
                 new_managers.add(room_data['owner'])
-            manager_group = room_attrs.get('manager-group')
             if manager_group is not None:
                 group = GroupProxy(manager_group, provider='cern-ldap')
                 if group.group is None:
--- a/foundationsync/indico_foundationsync.py
+++ b/foundationsync/indico_foundationsync.py
@@ -118,6 +118,7 @@ class FoundationSync(object):
         data['reservations_need_confirmation'] = raw_data['BOOKINGS_NEED_CONFIRMATION'] != 'N'
 
         is_reservable = raw_data['IS_RESERVABLE'] != 'N'
+        data['protection_mode'] = ProtectionMode.public if is_reservable else ProtectionMode.protected
         site = raw_data.get('SITE')
         site_map = {'MEYR': 'Meyrin', 'PREV': 'Prevessin'}
         data['site'] = site_map.get(site, site)
@@ -127,13 +128,14 @@ class FoundationSync(object):
             data['latitude'] = building_coordinates['latitude']
             data['longitude'] = building_coordinates['longitude']
 
-        return data, is_reservable, email_warning
+        return data, email_warning
 
     def _prepare_row(self, row, cursor):
         return dict(zip([d[0] for d in cursor.description], row))
 
     def _update_room(self, room, room_data, room_attrs):
         room.is_active = True
+        room.is_reservable = True
         for k, v in room_data.iteritems():
             if getattr(room, k) != v:
                 setattr(room, k, v)
@@ -208,7 +210,7 @@ class FoundationSync(object):
             room_id = data['ID']
 
             try:
-                room_data, is_reservable, email_warning = self._parse_room_data(data, coordinates, room_id)
+                room_data, email_warning = self._parse_room_data(data, coordinates, room_id)
                 room_attrs = self._get_room_attrs(data)
                 self._logger.debug("Fetched data for room with id='%s'", room_id)
             except SkipRoom as e:
@@ -253,7 +255,6 @@ class FoundationSync(object):
                 room.update_principal(principal, full_access=False)
             for principal in new_managers - current_managers:
                 room.update_principal(principal, full_access=True)
-            room.protection_mode = ProtectionMode.public if is_reservable else ProtectionMode.protected
 
             self._logger.info("Updated room '%s' information", room_id)
             foundation_rooms.append(room)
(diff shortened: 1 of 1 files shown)


foundationsync/indico_foundationsync.py
```diff
@@ -80,6 +84,10 @@ def connect_to_foundation(self):
     def _get_room_attrs(self, raw_data):
         return {'manager-group': raw_data.get('EMAIL_LIST')}
 
+    def _fix(self, s):
```
pferreir: `_fix` seems like too broad of a name. What do you think of `_fix_html`? Or event `_html_to_markdown`, since that's what the function actually does.

foundationsync/indico_foundationsync.py
```diff
@@ -104,9 +112,9 @@ def _parse_room_data(self, raw_data, coordinates, room_id):
         data['capacity'] = int(raw_data['CAPACITY']) if raw_data['CAPACITY'] else None
         data['surface_area'] = int(raw_data['SURFACE']) if raw_data['SURFACE'] else None
         data['division'] = raw_data.get('DEPARTMENT')
-        data['telephone'] = raw_data.get('TELEPHONE')
-        data['key_location'] = raw_data.get('WHERE_IS_KEY')
-        data['comments'] = raw_data.get('COMMENTS')
+        data['telephone'] = raw_data.get('TELEPHONE') if raw_data.get('TELEPHONE') else ''
```
pferreir: This a normal `dict` right? Then you could do `raw_data.get('TELEPHONE', '')`

foundationsync/indico_foundationsync.py
```diff
@@ -104,9 +112,9 @@ def _parse_room_data(self, raw_data, coordinates, room_id):
...
+        data['key_location'] = self._fix(str(raw_data.get('WHERE_IS_KEY'))) if raw_data.get('WHERE_IS_KEY') else ''
```
pferreir: This `str(...)` looks like it's gonna blow if there are non-ASCII chars. I'm not 100% sure since I don't know what the underlying cxOracle implementation is returning, but I'd just double check that it works ok with accented characters.

foundationsync/indico_foundationsync.py
```diff
@@ -231,8 +239,17 @@ def fetch_rooms(self, connection, room_name=None):
 
             # Update room data
             room = self._update_room(room, room_data, room_attrs)
+            for manager in room.get_manager_list():
+                room.update_principal(manager, full_access=False)
```
pferreir: Maybe I'm misunderstanding, but doesn't this do basically nothing?

foundationsync/indico_foundationsync.py
```diff
@@ -231,8 +239,17 @@ def fetch_rooms(self, connection, room_name=None):
...
+            if room_data.get('owner') is not None:
+                room.update_principal(room_data['owner'], full_access=True)
+            manager_group = room_attrs.get('manager-group')
+            if manager_group is not None:
+                group = GroupProxy.get_named_default_group(manager_group)
```
pferreir: The docstring of this method says it's deprecated.

foundationsync/indico_foundationsync.py
ThiefMaster: might be NULL in the DB - in that case `raw_data.get('TELEPHONE') or ''` should work though

foundationsync/indico_foundationsync.py
ThiefMaster: AFAIK we configured the oracle client to return unicode objects somewhere near the top of the file, so `str()` seems likely to break in that case

foundationsync/indico_foundationsync.py
pferreir: Oh right. How is the DB NULL represented? Is it falsy?

foundationsync/indico_foundationsync.py
ThiefMaster: Add a plugin setting to select the identity provider used (since gesloc just gives us group names) or hardcode `cern-ldap` for the provider name (setting is cleaner)

pferreir: I'd just hardcode it. It's a CERN plugin anyway. You cannot be more CERN than Foundation.

foundationsync/indico_foundationsync.py
ThiefMaster: `None` i guess

foundationsync/indico_foundationsync.py
pferreir: OK, got it. Let's add a comment.

foundationsync/indico_foundationsync.py
```diff
@@ -104,9 +112,9 @@ def _parse_room_data(self, raw_data, coordinates, room_id):
         data['capacity'] = int(raw_data['CAPACITY']) if raw_data['CAPACITY'] else None
         data['surface_area'] = int(raw_data['SURFACE']) if raw_data['SURFACE'] else None
         data['division'] = raw_data.get('DEPARTMENT')
-        data['telephone'] = raw_data.get('TELEPHONE')
-        data['key_location'] = raw_data.get('WHERE_IS_KEY')
-        data['comments'] = raw_data.get('COMMENTS')
+        data['telephone'] = raw_data.get('TELEPHONE') or ''
+        data['key_location'] = self._html_to_markdown(raw_data.get('WHERE_IS_KEY')) if raw_data.get('WHERE_IS_KEY') else ''
+        data['comments'] = self._html_to_markdown(raw_data.get('COMMENTS')) if raw_data.get('COMMENTS') else ''
         data['is_reservable'] = raw_data['IS_RESERVABLE'] != 'N'
```
ThiefMaster: This one changed: Rooms that are not marked as reservable in Indico are now also not reservable by its managers. So this column shouldn't be updated anymore (it should already be True)

foundationsync/indico_foundationsync.py
```diff
@@ -231,8 +239,19 @@ def fetch_rooms(self, connection, room_name=None):
...
+            for manager in room.get_manager_list():
+                room.update_principal(manager, full_access=False)
+            if room_data.get('owner') is not None:
+                room.update_principal(room_data['owner'], full_access=True)
+            manager_group = room_attrs.get('manager-group')
+            if manager_group is not None:
+                group = GroupProxy(manager_group, provider='cern-ldap')
+                room.update_principal(group, full_access=True)
+            if room_data['is_reservable']:
+                room.default_protection_mode = ProtectionMode.protected
```
ThiefMaster: This doesn't seem correct, the room should be protected if it's NOT (publicly) reservable, and public otherwise.

foundationsync/indico_foundationsync.py
```diff
@@ -231,8 +239,19 @@ def fetch_rooms(self, connection, room_name=None):
 
             # Update room data
             room = self._update_room(room, room_data, room_attrs)
+            # Remove all existing full_access acl entries before adding the new ones
+            # otherwise changing the room owner/manager-group wont remove the old one
+            for manager in room.get_manager_list():
+                room.update_principal(manager, full_access=False)
```
ThiefMaster: do not remove the current room owner here if there is no new owner, otherwise the current owner loses access even though he's still listed as the owner

foundationsync/indico_foundationsync.py
```diff
@@ -231,8 +239,19 @@ def fetch_rooms(self, connection, room_name=None):
...
+            if room_data.get('owner') is not None:
+                room.update_principal(room_data['owner'], full_access=True)
+            manager_group = room_attrs.get('manager-group')
+            if manager_group is not None:
+                group = GroupProxy(manager_group, provider='cern-ldap')
```
ThiefMaster: let's log a warning if `group.group is None` (that means the group does not exist in ldap)

foundationsync/indico_foundationsync.py
```diff
@@ -231,8 +239,19 @@ def fetch_rooms(self, connection, room_name=None):
...
+                room.update_principal(group, full_access=True)
+            if not is_reservable:
+                room.default_protection_mode = ProtectionMode.protected
```
ThiefMaster: `default_protection_mode` isn't assignable on the instance level... also, there's the else part missing:

    room.protection_mode = ProtectionMode.public if is_reservable else ProtectionMode.protected

foundationsync/indico_foundationsync.py
```diff
@@ -104,12 +112,12 @@ def _parse_room_data(self, raw_data, coordinates, room_id):
         data['capacity'] = int(raw_data['CAPACITY']) if raw_data['CAPACITY'] else None
         data['surface_area'] = int(raw_data['SURFACE']) if raw_data['SURFACE'] else None
         data['division'] = raw_data.get('DEPARTMENT')
-        data['telephone'] = raw_data.get('TELEPHONE')
-        data['key_location'] = raw_data.get('WHERE_IS_KEY')
-        data['comments'] = raw_data.get('COMMENTS')
-        data['is_reservable'] = raw_data['IS_RESERVABLE'] != 'N'
+        data['telephone'] = raw_data.get('TELEPHONE') or ''
+        data['key_location'] = self._html_to_markdown(raw_data.get('WHERE_IS_KEY')) if raw_data.get('WHERE_IS_KEY') else ''
```
mic4ael: This line exceeds the limit of 120 characters. Maybe use normal `if` statement?

foundationsync/indico_foundationsync.py
```diff
@@ -119,7 +127,7 @@ def _parse_room_data(self, raw_data, coordinates, room_id):
             data['latitude'] = building_coordinates['latitude']
             data['longitude'] = building_coordinates['longitude']
 
-        return data, email_warning
+        return data, is_reservable, email_warning
```
ThiefMaster: why not just put the correct `protection_mode` in the `data` in this function? then you don't need extra code to set it later since we copy all the data to the room in `_update_room`

PR #56, opened 2019-06-14
Update the workflow of creating room assistance request

mic4ael:
This is still a Draft since I need to just make sure that everything works fine. And apart from that I need to create proper commits with proper commit messages ;)

SQL command to update the request data:

```sql

UPDATE events.requests
SET data = data - 'start_dt' || json_build_object('occurrences', json_build_array(data->>'start_dt'))::JSONB
WHERE type = 'room-assistance' AND data ? 'start_dt';
```

### DIFF
--- a/room_assistance/indico_room_assistance/client/style.scss
+++ b/room_assistance/indico_room_assistance/client/style.scss
@@ -1,0 +1,32 @@
+// This file is part of the CERN Indico plugins.
+// Copyright (C) 2014 - 2019 CERN
+//
+// The CERN Indico plugins are free software; you can redistribute
+// them and/or modify them under the terms of the MIT License; see
+// the LICENSE file for more details.
+
+.assistance-occurrences-widget {
+  display: flex;
+  flex-direction: column;
+
+  .occurrence + .occurrence {
+    margin-top: 5px;
+  }
+
+  input ~ input {
+    margin-right: 1em;
+  }
+
+  .occurrence {
+    align-items: center;
+    display: inline-flex;
+
+    .datepicker {
+      width: 40% !important; // sass-lint:disable-line no-important
+    }
+
+    .timepicker {
+      width: 20% !important; // sass-lint:disable-line no-important
+    }
+  }
+}
--- a/room_assistance/indico_room_assistance/client/style.scss
+++ b/room_assistance/indico_room_assistance/client/style.scss
@@ -6,27 +6,27 @@
 // the LICENSE file for more details.
 
 .assistance-occurrences-widget {
-  display: flex;
-  flex-direction: column;
+    display: flex;
+    flex-direction: column;
 
-  .occurrence + .occurrence {
-    margin-top: 5px;
-  }
-
-  input ~ input {
-    margin-right: 1em;
-  }
-
-  .occurrence {
-    align-items: center;
-    display: inline-flex;
-
-    .datepicker {
-      width: 40% !important; // sass-lint:disable-line no-important
+    .occurrence + .occurrence {
+        margin-top: 5px;
     }
 
-    .timepicker {
-      width: 20% !important; // sass-lint:disable-line no-important
+    input ~ input {
+        margin-right: 1em;
     }
-  }
+
+    .occurrence {
+        align-items: center;
+        display: inline-flex;
+
+        .datepicker {
+            width: 40% !important; // sass-lint:disable-line no-important
+        }
+
+        .timepicker {
+            width: 20% !important; // sass-lint:disable-line no-important
+        }
+    }
 }
--- a/room_assistance/indico_room_assistance/client/style.scss
+++ b/room_assistance/indico_room_assistance/client/style.scss
@@ -22,11 +22,11 @@
         display: inline-flex;
 
         .datepicker {
-            width: 40% !important; // sass-lint:disable-line no-important
+            width: 40% !important;
         }
 
         .timepicker {
-            width: 20% !important; // sass-lint:disable-line no-important
+            display: flex;
         }
     }
 }
--- a/room_assistance/indico_room_assistance/forms.py
+++ b/room_assistance/indico_room_assistance/forms.py
@@ -9,6 +9,7 @@ from __future__ import unicode_literals
 
 from datetime import date, time, timedelta
 
+import dateutil.parser
 from flask import request, session
 from wtforms import SelectField
 from wtforms.fields.html5 import IntegerField
@@ -16,10 +17,10 @@ from wtforms.fields.simple import TextAreaField
 from wtforms.validators import DataRequired, NumberRange, Optional, ValidationError
 
 from indico.modules.events.requests import RequestFormBase
-from indico.util.date_time import now_utc
 from indico.web.forms.base import IndicoForm, generated_data
-from indico.web.forms.fields import IndicoDateField, IndicoDateTimeField
-from indico.web.forms.validators import DateTimeRange, Exclusive
+from indico.web.forms.fields import IndicoDateField, JSONField
+from indico.web.forms.validators import Exclusive
+from indico.web.forms.widgets import JinjaWidget
 
 from indico_room_assistance import _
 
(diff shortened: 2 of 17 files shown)


room_assistance/indico_room_assistance/client/style.scss
```diff
@@ -0,0 +1,32 @@
+// This file is part of the CERN Indico plugins.
+// Copyright (C) 2014 - 2019 CERN
+//
+// The CERN Indico plugins are free software; you can redistribute
+// them and/or modify them under the terms of the MIT License; see
+// the LICENSE file for more details.
+
+.assistance-occurrences-widget {
+  display: flex;
```
ThiefMaster: we only use 2-space indentation for JS ;)

room_assistance/indico_room_assistance/forms.py
```diff
@@ -7,37 +7,65 @@
...
+
+        dts = []
+        tzinfo = pytz.timezone(config.DEFAULT_TIMEZONE)
+        for req_date, req_time in self.data.iteritems():
+            dt = tzinfo.localize(dateutil.parser.parse('{} {}'.format(req_date, req_time)))
+            dts.append(dt)
+        self.data = dts
+
+    def _value(self):
+        return {dt.date().isoformat(): dt.time().strftime('%H:%M') for dt in self.data} if self.data else {}
```
ThiefMaster: I don't think the `.time()` is needed here

room_assistance/indico_room_assistance/forms.py
```diff
@@ -7,37 +7,65 @@
...
-            raise ValidationError('Specified datetime is not within the working hours')
+    def __init__(self, *args, **kwargs):
+        super(RoomAssistanceRequestForm, self).__init__(*args, **kwargs)
+        self.occurrences.event = kwargs['event']
+
+    def validate_occurrences(self, field):
+        for req_dt in field.data:
+            is_in_working_hours = any(period[0] <= req_dt.time() <= period[1] for period in WORKING_TIME_PERIODS)
+            if not is_in_working_hours:
+                raise ValidationError('One of the specified times is not within the working hours')
```
ThiefMaster: i18n

room_assistance/indico_room_assistance/templates/assistance_request_occurrences.html
```diff
@@ -0,0 +1,67 @@
...
+        const $field = $('#{{ field.id }}');
+        const $occurrences = $('#{{ field.id }}-container .occurrence');
+
+        $occurrences
+            .find('.datepicker')
+            .datepicker({
+                dateFormat: 'dd/mm/yy',
+                disabled: true
+            })
+            .end()
```
ThiefMaster: can't we just start a new statement here? I don't think jumping around in the selector stack makes code particularly readable

mic4ael: We can, but at least to me using `end` makes this code a little bit more concise and readable.

room_assistance/indico_room_assistance/client/style.scss
mic4ael: Ah, now I see why it seemed off to me as well. Sass-lint reported that I should be using 2-space  indentation which is the default I guess. We are not using our `.sass-lint.yml` file when running `sass-lint` over plugins.

room_assistance/indico_room_assistance/util.py
```diff
@@ -13,3 +13,14 @@ def is_room_assistance_support(user):
...
     return RoomAssistancePlugin.settings.acls.contains_user('room_assistance_support', user)
+
+
+def event_has_room_with_support_attached(event):
+    from indico_room_assistance.plugin import RoomAssistancePlugin
+    return event.room in RoomAssistancePlugin.settings.get('rooms_with_assistance')
+
+
+def can_request_assistance_for_event(event):
+    has_room_assistance_request = event.requests.filter_by(type='room-assistance').has_rows()
```
ThiefMaster: Are you sure you want to take withdrawn/rejected requests into account here?!

room_assistance/indico_room_assistance/templates/request_list.html
```diff
@@ -39,7 +39,16 @@
...
     <tbody>
         <tr class="result-group-entry">
-            <td>{{ request.event.room.full_name }}</td>
+            <td>
+                <span>
+                    {{ request.event.room.full_name }}
+                </span>
+                {% if not event_has_supported_room(request.event) %}
+                    <i class="icon-warning"
+                       title="{% trans %}The event location is not in the list of supported rooms{% endtrans %}">
```
ThiefMaster: Maybe something like "Assistance is not available for the selected room"?

room_assistance/indico_room_assistance/templates/emails/creation_email_to_assistance.html
```diff
@@ -11,7 +11,7 @@
         {{ requested_by.full_name }} requested assistance to setup {{ event.room.full_name }} before the start of the event.
     </p>
     <div>
-        <strong>Event:</strong> {{ event.title }}<br>
+        <strong>Event:</strong> {{ event.title }} ({{ event.external_url }})<br>
```
ThiefMaster: It's an HTML email - why not use an HTML link as well? (could just make the title a link)

room_assistance/indico_room_assistance/util.py
ThiefMaster: `Request.find_latest_for_event()` might be useful here.

room_assistance/indico_room_assistance/templates/emails/creation_email_to_assistance.html
mic4ael: Oh, true.

room_assistance/indico_room_assistance/util.py
mic4ael: `Request.find_latest_for_event` is not very useful here if I want to filter by `Request`'s state.

ThiefMaster: well you get a single one and then check the state if it's not none. but yeah if you only care about the accepted one then querying manually is fine as well

room_assistance/indico_room_assistance/plugin.py
```diff
@@ -91,3 +96,27 @@ def _extend_services_menu(self, reservation, **kwargs):
...
+        changes = kwargs['changes']
+        if not changes.viewkeys() & {'location_data', 'start_dt', 'end_dt'}:
+            return
+
+        request = Request.find_latest_for_event(event, RoomAssistanceRequest.name)
+        if not request or request.state != RequestState.accepted:
+            return
+
+        if 'location_data' in changes and not event_has_room_with_support_attached(event):
+            flash(_("The new event location is not in the list of the rooms supported by the room assistance team. "
```
ThiefMaster: shouldn't we auto-reject the request in this case?

room_assistance/indico_room_assistance/util.py
```diff
@@ -7,9 +7,27 @@
...
+
+
+def event_has_room_with_support_attached(event):
+    from indico_room_assistance.plugin import RoomAssistancePlugin
+    return event.room in RoomAssistancePlugin.settings.get('rooms_with_assistance')
+
+
+def can_request_assistance_for_event(event):
+    has_room_assistance_request = (event.requests
+                                   .filter(Request.type == RoomAssistanceRequest.name,
```
ThiefMaster: nothing wrong with this, but `filter_by(type=.., state=...)` would be more concise here.

room_assistance/indico_room_assistance/plugin.py
mic4ael: We don't reject it for example in case of the `vc-assistance` request so I thought that it should be the same.

ThiefMaster: hmm i checked in audiovisual and there we do.. maybe we should reject in both cases? if you end up in a state where you wouldn't be able to send a request i think it makes sense if the event managers are clearly notified (the rejection triggers an email to all managers)

mic4ael: Ok, then maybe I will do the same in room assistance.

room_assistance/indico_room_assistance/client/style.scss
```diff
@@ -0,0 +1,32 @@
...
+    input ~ input {
+        margin-right: 1em;
+    }
+
+    .occurrence {
+        align-items: center;
+        display: inline-flex;
+
+        .datepicker {
+            width: 40% !important; // sass-lint:disable-line no-important
```
ThiefMaster: the sass-line rule shouldn't be needed, `no-important` isn't enabled in our config

room_assistance/indico_room_assistance/plugin.py
```diff
@@ -68,19 +75,18 @@ class RoomAssistancePlugin(IndicoPlugin):
 
     def init(self):
         super(RoomAssistancePlugin, self).init()
+        self.inject_bundle('main.css', WPRequestsEventManagement)
```
ThiefMaster:
```python
        self.inject_bundle('main.css', WPRequestsEventManagement, subclasses=False,
                           condition=lambda: request.view_args.get('type') == RoomAssistanceRequest.name)
```

room_assistance/indico_room_assistance/plugin.py
```diff
@@ -91,3 +97,29 @@ def _extend_services_menu(self, reservation, **kwargs):
...
+            data = dict(request.data, comment=render_plugin_template('auto_reject_no_supported_room.txt'))
+            request.definition.reject(request, data, User.get_system_user())
+            flash(_("The new event location is not in the list of the rooms supported by the room assistance team. "
+                    "Room assistance request has been rejected and support will not be provided."), 'warning')
+        if changes.viewkeys() & {'start_dt', 'end_dt'}:
+            tz = pytz.timezone(config.DEFAULT_TIMEZONE)
+            req_dates = {dateutil.parser.parse(occ).astimezone(tz).date() for occ in request.data['occurrences']}
+            event_dates = set(event.iter_days())
+            old_dates = req_dates - event_dates
+            if old_dates:
```
ThiefMaster: If there is no overlap between old and new dates I think we should reject the request.

room_assistance/indico_room_assistance/client/style.scss
```diff
@@ -0,0 +1,32 @@
...
+    .occurrence {
+        align-items: center;
+        display: inline-flex;
+
+        .datepicker {
+            width: 40% !important;
+        }
+
+        .timepicker {
+            width: 20% !important;
```
mvidalgarcia: I think you're trying to hide the `PM/AM` in the time picker? If that's the case what's the reason behind? this is causing problems in Chrome.
image

mic4ael: Nope I am not trying to hide it. That is a good catch I didn't notice that since I couldn't even see the P like in your case.

mic4ael: Hm, IT is a bit strange since I thought I was using 24h clock format there. Are you using different locale or sth?

mvidalgarcia: `en-US` AFAIK, `es-ES` as second language. In both Chrome and Firefox I get 12h clock.

mic4ael: Increased the width of the time field.

meluru:
I checked out your branch and I think there's still AM/PM problem.

image

mic4ael:
> I checked out your branch and I think there's still AM/PM problem.
> 
> image

Did you rebuild assets etc?

mic4ael:
Ok, I see. I will check this out.

meluru:
Yes, I checked twice because I also thought this has been solved.

mic4ael:
Yeah, fixed now.

room_assistance/indico_room_assistance/client/style.scss
mvidalgarcia: Looks good now! :+1:

PR #63, opened 2019-07-31
RoomAssistance: Require user to specify time for room assistance request

mic4ael:


room_assistance/indico_room_assistance/forms.py
```diff
@@ -39,12 +39,24 @@ def process_formdata(self, valuelist):
...
         self.data = dts
 
     def _value(self):
-        return {dt.date().isoformat(): dt.strftime('%H:%M') for dt in self.data} if self.data else {}
+        if not self.data:
+            return {}
+
+        data = {}
+        for dt in self.data:
+            if type(dt) is datetime:
```
ThiefMaster: `isinstance()`?

room_assistance/indico_room_assistance/forms.py
```diff
@@ -39,12 +39,24 @@ def process_formdata(self, valuelist):
...
+                data[dt.date().isoformat()] = dt.strftime('%H:%M')
+            elif type(dt) is date:
```
ThiefMaster: `isinstance()`

room_assistance/indico_room_assistance/forms.py
```diff
@@ -62,6 +74,8 @@ def __init__(self, *args, **kwargs):
 
     def validate_occurrences(self, field):
         for req_dt in field.data:
+            if type(req_dt) is date:
```
ThiefMaster: `isinstance()`

ThiefMaster: and i'd rather do `not isinstance(req_dt, datetime)`

room_assistance/indico_room_assistance/forms.py
mic4ael: I know but I wanted to be clear here since `isinstance(datetime.now(), datetime.date)` returns True.

room_assistance/indico_room_assistance/forms.py
mic4ael: The same as my response above.

room_assistance/indico_room_assistance/forms.py
ThiefMaster: WTF, why is a datetime a date subclass (but not a time subclass)?!

room_assistance/indico_room_assistance/forms.py
ThiefMaster: I think here you *can* safely  use the isinstance check since you really want a `datetime`, and that one is not a subclass of date nor time

mic4ael:
> Please also bump the version by +0.0.1

Should it be `2.2.1` now?

ThiefMaster:
Yes

PR #65, opened 2019-09-03
FoundationSync: sync with AIS roles and remove equipment sync

pferreir:
Non-synced fields (they are all moved to Indico):
 * "Can be reserved"
 * "Requires confirmation"
 * Equipments
 * Management e-group

### DIFF
--- a/foundationsync/indico_foundationsync/plugin.py
+++ b/foundationsync/indico_foundationsync/plugin.py
@@ -64,7 +64,7 @@ class FoundationSyncPlugin(IndicoPlugin):
         return foundationsync
 
 
-@celery.periodic_task(run_every=crontab(minute='0', hour='8'))
+@celery.periodic_task(run_every=crontab(minute='0'))
 def scheduled_update(room_name=None):
     db_name = FoundationSyncPlugin.settings.get('connection_string')
     if not db_name:
--- a/foundationsync/indico_foundationsync/plugin.py
+++ b/foundationsync/indico_foundationsync/plugin.py
@@ -16,15 +16,24 @@ from logging import StreamHandler
 
 import click
 from celery.schedules import crontab
+from flask import g, session
+from webargs import flaskparser
 from wtforms import StringField
 
 from indico.cli.core import cli_command
 from indico.core import signals
 from indico.core.celery import celery
 from indico.core.plugins import IndicoPlugin
+from indico.modules.rb.schemas import RoomUpdateArgsSchema
+from indico.util.i18n import _
 from indico.web.forms.base import IndicoForm
 
 from indico_foundationsync.sync import FoundationSync
+
+
+SYNCED_FIELDS = {'building', 'floor', 'number', 'verbose_name', 'site', 'key_location', 'capacity', 'surface_area',
+                 'telephone', 'division', 'comments', 'latitude', 'longitude', 'owner'}
+BLOCKED_FIELDS = {'protection_mode'}
 
 
 class SettingsForm(IndicoForm):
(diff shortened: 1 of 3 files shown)


foundationsync/indico_foundationsync/sync.py
```diff
@@ -46,6 +41,15 @@ def OutputTypeHandler(cursor, name, defaultType, size, precision, scale):
         return cursor.var(unicode, size, cursor.arraysize)
 
 
+def _get_room_role_map(connection, logger):
+    roles = defaultdict(set)
+    cursor = connection.cursor()
+    cursor.execute('SELECT BUILDING, FLOOR, ROOM_NUMBER, EMAIL FROM aispub.app_indico_space_managers')
+    for row in cursor:
+        roles[row[:3]].add(row[3])
```
ThiefMaster: `row['EMAIL']`?

i don't think using positional indexes for query results with more than one row is great, but i have to admit that it's convenient for the first 3 rows ;)

foundationsync/indico_foundationsync/sync.py
```diff
@@ -125,10 +131,35 @@ def _update_room(self, room, room_data):
...
                 setattr(room, k, v)
         db.session.flush()
 
+    def _update_managers(self, room, room_data, room_role_map):
+        new_managers = {room.owner}
+        manager_group = room_data.get('manager_group')
+
+        # add managers from e-group
+        if manager_group is not None:
+            group = GroupProxy(manager_group.strip(), provider='cern-ldap')
```
ThiefMaster: can we also strip `@cern.ch`? right now there's at least one room in the DB that has this (which results in some unnecessary noise in the sentry logs)

foundationsync/indico_foundationsync/sync.py
```diff
@@ -125,10 +131,35 @@ def _update_room(self, room, room_data):
...
+            if group.group is None:
+                self._logger.warning("Group '%s' does not exist in LDAP", manager_group)
+            new_managers.add(group)
+
+        # add managers from aisroles (DKMs + DKAs)
+        new_managers |= {get_user_by_email(email, create_pending=True)
+                         for email in room_role_map[(room.building, room.floor, room.number)]}
```
ThiefMaster: can we end up with a room with no entry in the role map?

pferreir: No, since the owner will always be there (otherwise we won't update it).

foundationsync/indico_foundationsync/sync.py
pferreir: Since I enumerate them above, I don't see that as a big deal. But yeah, `EMAIL` is better written as text.

foundationsync/indico_foundationsync/plugin.py
```diff
@@ -43,11 +51,34 @@ class FoundationSyncPlugin(IndicoPlugin):
...
         from indico_foundationsync.blueprint import blueprint
         return blueprint
 
+    def _check_forbidden_fields(self, sender, data, **kwargs):
+        """Check that no one is trying to edit fields that come from Gesloc."""
+        sync_collisions = set(data) & SYNCED_FIELDS
+        blocked_collisions = set(data) & BLOCKED_FIELDS
+        messages = {}
+
+        if sync_collisions:
```
mic4ael: I guess this `if` is not needed at all.

```suggestion
            messages.update({
                k: [_('This field is managed by Gesloc. You cannot change it from here.')] for k in sync_collisions
            })

```

ThiefMaster:
Shouldn't this be based on `2.2-maintenance` instead of `master`?

foundationsync/indico_foundationsync/plugin.py
```diff
@@ -16,17 +16,25 @@
...
+from indico.util.i18n import _
 from indico.web.forms.base import IndicoForm
 
 from indico_foundationsync.sync import FoundationSync
 
 
+SYNCED_FIELDS = {'building', 'floor', 'number', 'verbose_name', 'site', 'key_location', 'capacity', 'surface_area',
+                 'telephone', 'division', 'comments', 'latitude', 'longitude', 'owner'}
+BLOCKED_FIELDS = {'protection_mode'}
+
```
ThiefMaster: there should be 2 blank lines here

foundationsync/indico_foundationsync/plugin.py
```diff
@@ -43,11 +52,39 @@ class FoundationSyncPlugin(IndicoPlugin):
...
+        """Check that no one is trying to edit fields that come from Gesloc."""
+        sync_collisions = set(data) & SYNCED_FIELDS
+        blocked_collisions = set(data) & BLOCKED_FIELDS
+        messages = {}
+
+        messages.update({
+            k: [_('This field is managed by Gesloc. You cannot change it from here.')] for k in sync_collisions
+        })
+
+        if blocked_collisions and not session.user.is_admin:
```
ThiefMaster: `rb_is_admin(session.user)`

PR #69, opened 2019-10-09
VC Assistance: add endpoint to export bookings

pferreir:
A simple endpoint (`/export/room_vc_list/xxx/y-zz.json`) which exports all room bookings for a room in a given time span, as well as any associated events and VC rooms.

### DIFF
--- a/vc_assistance/indico_vc_assistance/api.py
+++ b/vc_assistance/indico_vc_assistance/api.py
@@ -7,10 +7,17 @@
 
 from __future__ import unicode_literals
 
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 import icalendar
+from sqlalchemy.orm import joinedload, noload
+from werkzeug.exceptions import NotFound
 
+from indico.core.db import db
+from indico.modules.events.models.events import Event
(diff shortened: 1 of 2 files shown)


vc_assistance/indico_vc_assistance/api.py
```diff
@@ -106,3 +113,82 @@ def _ical_serialize_vc_alarm(record):
...
+    def _getParams(self):
+        super(RoomVCListHook, self)._getParams()
+        self.room = Room.query.filter(
+            ~Room.is_deleted,
+            Room.building == self._pathParams['building'],
+            Room.floor == self._pathParams['floor'],
+            Room.number == self._pathParams['number']
+        ).one_or_none()
+        if not self.room:
+            raise NotFound
```
ThiefMaster: sounds like you can just use `.one()` above

vc_assistance/indico_vc_assistance/api.py
```diff
@@ -106,3 +113,82 @@ def _ical_serialize_vc_alarm(record):
...
+
+    def _has_access(self, user):
+        return is_vc_support(user)
+
+    def export_room_vc_list(self, user):
+
```
ThiefMaster: :space_invader:

PR #73, opened 2019-12-11
Improve wording in room assistance plugin

pferreir:
As requested by @tgbaron

### DIFF
--- a/room_assistance/indico_room_assistance/templates/event_request_details.html
+++ b/room_assistance/indico_room_assistance/templates/event_request_details.html
@@ -32,7 +32,8 @@
     {% else %}
         {%- call message_box('info') -%}
             {% trans room_name=event.room.full_name %}
-                Assistance will be provided for the room <strong>{{ room_name }}</strong>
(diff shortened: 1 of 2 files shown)


room_assistance/indico_room_assistance/templates/event_request_details.html
```diff
@@ -32,7 +32,8 @@
     {% else %}
         {%- call message_box('info') -%}
             {% trans room_name=event.room.full_name %}
-                Assistance will be provided for the room <strong>{{ room_name }}</strong>
+                A support technician can come to room <strong>{{ room_name }}</strong> in order to help you with the
```
ThiefMaster: `to <room name>` or `to the room <room name>`?

pferreir: Does "someone will come to room 28/R-020" sound bad to you? Not sure, I find that the article is not needed, but it would be nice to have the opinion of a native.

ThiefMaster: indeed, doesn't sound bad. I had something like "to room main auditorium" in mind, which does sound weird

PR #121, opened 2022-11-02
Add CloudConvert support

tomasr8:
Adds experimental CloudConvert support to the conversion plugin.
Feedback welcome!

### How to test this:
- Make an account on CloudConvert and create an API key (you can also use the sandbox API which doesn't count towards the limit: https://cloudconvert.com/api/v2#overview, but then you can only upload limited files based on an md5 hash)
- Enable CloudConvert instead of Doconverter and set the API key in the Conversion plugin setttings.
- Make sure the webhook (`/conversion/cloudconvert/finished`) is reachable by cloudconvert (use ngrok for reverse proxy if needed)
- Upload a file with a supported extension (e.g. `docx`) and toggle `Convert to PDF`
- Wait & Pray

### DIFF
--- a/conversion/indico_conversion/blueprint.py
+++ b/conversion/indico_conversion/blueprint.py
@@ -7,9 +7,12 @@
 
 from indico.core.plugins import IndicoPluginBlueprint
 
-from indico_conversion.conversion import RHConversionCheck, RHConversionFinished
+from indico_conversion.conversion import RHCloudConvertFinished, RHConversionCheck, RHDoconverterFinished
 
 
 blueprint = IndicoPluginBlueprint('conversion', __name__)
-blueprint.add_url_rule('/conversion/finished', 'callback', RHConversionFinished, methods=('POST',))
+blueprint.add_url_rule('/conversion/doconverter/finished', 'doconverter_callback',
+                       RHDoconverterFinished, methods=('POST',))
+blueprint.add_url_rule('/conversion/cloudconvert/finished', 'cloudconvert_callback',
+                       RHCloudConvertFinished, methods=('POST',))
 blueprint.add_url_rule('/conversion/check', 'check', RHConversionCheck)
--- a/conversion/indico_conversion/conversion.py
+++ b/conversion/indico_conversion/conversion.py
@@ -127,7 +127,7 @@ def submit_attachment_cloudconvert(task, attachment):
     import_task = f'import-file-{attachment.id}'
     convert_task = f'convert-file-{attachment.id}'
     export_task = f'export-file-{attachment.id}'
-    signed_id = secure_serializer.dumps(str(attachment.id), salt='pdf-conversion')
+    signed_id = secure_serializer.dumps(attachment.id, salt='pdf-conversion')
 
     job_definition = {
         'tag': f'{attachment.file.id}__{signed_id}',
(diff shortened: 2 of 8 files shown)


conversion/indico_conversion/blueprint.py
```diff
@@ -7,9 +7,12 @@
 
 from indico.core.plugins import IndicoPluginBlueprint
 
-from indico_conversion.conversion import RHConversionCheck, RHConversionFinished
+from indico_conversion.conversion import RHConversionCheck, RHDoconverterFinished, RHCloudConvertFinished
 
 
 blueprint = IndicoPluginBlueprint('conversion', __name__)
-blueprint.add_url_rule('/conversion/finished', 'callback', RHConversionFinished, methods=('POST',))
+blueprint.add_url_rule('/conversion/doconverter/finished', 'doconverter-callback',
```
ThiefMaster, line 14: I think we use underscores in all other endpoint names (it's just in URLs where they are ugly)

conversion/indico_conversion/conversion.py
```diff
@@ -98,20 +102,80 @@ def _process(self):
...
+
+
+@celery.task(bind=True, max_retries=3)
+def submit_attachment_cloudconvert(task, attachment):
+    """Sends an attachment's file to the CloudConvert conversion service"""
+    from indico_conversion.plugin import ConversionPlugin
+    if ConversionPlugin.settings.get('maintenance'):
+        task.retry(countdown=900)
+
+    cloudconvert.configure(api_key=ConversionPlugin.settings.get('cloudconvert_api_key'), sandbox=True)
```
ThiefMaster, line 117: maybe `sandbox` could be a plugin setting as well?

ThiefMaster, line 117: using global state for this is ew. `configure` calls this:

```python
def set_config(options=None, **config):
    """Create new default api object with given configuration
    """
    global __client__
    __client__ = CloudConvertRestClient(options or {}, **config)
    return __client__
```

I'm not sure why they are doing such obscure things. I opened an issue cloudconvert/cloudconvert-python#26 about this...

conversion/indico_conversion/conversion.py
```diff
@@ -98,20 +102,80 @@ def _process(self):
...
+
+    job = cloudconvert.Job.create(payload={
+        "tag": str(attachment.id),
```
ThiefMaster, line 120: single quotes please ;)

conversion/indico_conversion/conversion.py
```diff
@@ -98,20 +102,80 @@ def _process(self):
...
+    """Callback to attach a converted file"""
+
+    CSRF_ENABLED = False
+
+    def _process(self):
+        from indico_conversion.plugin import ConversionPlugin
+
+        event = request.json['event']
+        if event == 'job.failed':
+            ConversionPlugin.logger.error(f'Cloudconvert conversion job failed: {event}')
```
ThiefMaster, line 165: `.error('whatever: %s', event)` so we get proper grouping on sentry.

Even though in this case passing `event` may not be that useful in the first place because it's always `job.failed`

conversion/indico_conversion/conversion.py
ThiefMaster, line 120: could we use signed data for this like in the doconverter `dirresponse`?

conversion/indico_conversion/conversion.py
```diff
@@ -131,3 +200,28 @@ def _process(self):
...
+    notify_threshold = ConversionPlugin.settings.get('notify_threshold')
+    notify_email = ConversionPlugin.settings.get('notify_email')
+
+    try:
+        credits = client.get_remaining_credits()
+    except requests.RequestException as err:
+        ConversionPlugin.logger.info('Could not fetch the remaining CloudConvert credits: %s', err)
+        task.retry(countdown=60)
+
+    if notify_threshold is not None and credits <= notify_threshold:
```
ThiefMaster, line 220: Does it make sense to even run the task (and send the request) when the threshold is None?

tomasr8: good point, I'd say no

PR #166, opened 2024-06-12
[i18n-demo] Disable cloning when user not logged in

tomasr8:
We need the user to be logged in to create a subcategory for them.
If a user is not logged in, the button is disabled with a tooltip telling them to log in.

### DIFF
--- a/i18n_demo/indico_i18n_demo/controllers.py
+++ b/i18n_demo/indico_i18n_demo/controllers.py
@@ -7,18 +7,18 @@
 
 from flask import flash, redirect, session
 from flask_pluginengine import current_plugin
-from werkzeug.exceptions import BadRequest, Forbidden
+from werkzeug.exceptions import BadRequest
 
 from indico.core.db.sqlalchemy.protection import ProtectionMode
 from indico.modules.categories.models.categories import Category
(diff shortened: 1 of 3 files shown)


i18n_demo/indico_i18n_demo/controllers.py
```diff
@@ -30,6 +30,9 @@ def _process(self):
         if not (category_id := current_plugin.settings.get('test_category_id')):
             raise BadRequest('No test category ID configured')
 
+        if not session.user:
+            raise Forbidden('You must be logged in to clone events')
```
ThiefMaster, line 34: Simply inherit your RH from `RHProtectedEventBase` instead of `RHEventBase`

i18n_demo/indico_i18n_demo/plugin.py
```diff
@@ -55,7 +55,7 @@ def _inject_clone_button(self, event, **kwargs):
             return
 
         if event.category != test_category and not event.category.is_descendant_of(test_category):
-            return render_plugin_template('clone_button.html', event=event)
+            return render_plugin_template('clone_button.html', event=event, disabled=not session.user)
```
ThiefMaster, line 58: `session.user` is also available inside any template

PR #168, opened 2024-06-20
Do not send emails when the user is not logged in

tomasr8:
If there's no `session.user` we do not send the emails.
The exception is verification emails to allow people to create local accounts.

### DIFF
--- a/i18n_demo/indico_i18n_demo/plugin.py
+++ b/i18n_demo/indico_i18n_demo/plugin.py
@@ -67,8 +67,15 @@ class I18nDemoPlugin(IndicoPlugin):
             # In those cases, we trust the users not to spam random people.
             return ret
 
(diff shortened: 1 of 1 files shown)


i18n_demo/indico_i18n_demo/plugin.py
```diff
@@ -67,8 +67,14 @@ def _intercept_make_email(self, sender, func, args, **kwargs):
             # In those cases, we trust the users not to spam random people.
             return ret
 
+        if session.get('register_verification_email_sent'):
+            # Let verification emails through
+            return ret
+
+        # If the user is logged in, redirect the email to them, otherwise do not send it
+        to = session.user.email if session.user else 'blackhole@indico.invalid'
```
ThiefMaster, line 75: Does this create bounces from the mail server?

PR #172, opened 2024-08-09
Add Typescript support in the (CERN) plugins

pferreir:
There's also a demo `RowLabel.tsx` just to get things started. Untested, now that I think of it.

TODO:
 * [x] Test that `RowLabel` is not broken, before merging


ThiefMaster:
RowLabel works fine, just tested it

PR #173, opened 2024-08-09
Zoom Rooms Plugin

pferreir:
This is the third instalment of https://github.com/indico/indico/pull/6475. See also https://github.com/indico/indico-plugins/pull/236

Changes:
 * RAVEM - move to SUI, in line with changes in core
 * Zoom Rooms - new plugin and tests, as well as some documentation (README)
 
 TODO:
 * [x] Replace `https://link/to/docs` with the true URL, once there is one. Or even better, add a plugin setting to set the URL.
 * [x] Fix failing unit test and add unit tests to CI
 * [x] Add some more docstrings


zoom_rooms/indico_zoom_rooms/models.py
```diff
@@ -0,0 +1,143 @@
...
+        JSONB(none_as_null=True),
+    )
+    #: Additional args to be sent with the request
+    extra_args: ExtraArgs = db.Column(
+        JSONB(none_as_null=True),
+    )
+
+    def __repr__(self):
+        action = ZoomRoomsAction(self.action).name
+        return f'<ZoomRoomsQueueEntry({self.id}, {self.entry_id}, {action})>'
```
ThiefMaster, line 99: there's a nice `format_repr` util for that

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,280 @@
...
+    ):
+        link_obj = obj.link_object if isinstance(obj, VCRoomEventAssociation) else obj
+
+        if not link_obj.room:
+            return
+
+        zr_id = get_zoom_room_id(link_obj.room)
+
+        # do not bother doing anything if the room is not ZR-enabled
+        if zr_id:
```
ThiefMaster, line 63:
```suggestion
        # do not bother doing anything if the room is not ZR-enabled
        if zr_id := get_zoom_room_id(link_obj.room):
```

zoom_rooms/indico_zoom_rooms/util.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if (
+        room
+        and (attributes := room.attributes)
+        and (
+            user_id_attr := attributes.filter(RoomAttribute.name == 'zoom-rooms-calendar-id')
+            .join(RoomAttribute)
+            .first()
+        )
+    ):
```
ThiefMaster, line 35: really? :D 

(not a fan of running queries inside an `if` unless they all fit in a single line)

ThiefMaster (commented):
I know it's still a draft but I had the PR open and couldn't resist having a look 🙈

ravem/indico_ravem/client/index.js
```diff
@@ -198,9 +198,7 @@
       const name = btn.data('roomName');
       const vcRoomName = btn.data('vcRoomName');
       const html = [
-        '<span class="',
-        states[newState].icon,
-        '"><strong style="margin-left: 0.4em;">{0}</strong></span>'.format(name),
+        `<i class="icon ${states[newState].icon}"></i> ${name}`
       ].join('');
```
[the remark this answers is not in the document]
pferreir: Indeed.

zoom_rooms/indico_zoom_rooms/util.py
pferreir: I wanted to avoid nesting ifs. TBH I find them equally readable. But I can change it, no strong opinion.

zoom_rooms/indico_zoom_rooms/util.py
```diff
@@ -0,0 +1,94 @@
...
+    return f'zoom_meeting:{user_id}@indico:event:{event_id}{sub_id}#{zoom_id}'
+
+
+def get_vc_room_associations(obj: Session | SessionBlock | Contribution | Event) -> t.Iterable[VCRoomEventAssociation]:
+    """
+    Get all VC Room Associations tied to the object plus those tied to nested objects inheriting the location from it.
+
+    :param obj: the object in question (`Session`, `SessionBlock`, `Contribution`, `Event`)
+    """
+    _flatten = lambda it: (assoc for assocs in it for assoc in assocs)
```
ThiefMaster, line 60: there's nothing in itertools for this?

zoom_rooms/indico_zoom_rooms/util.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # drill down into the object's dependencies to figure out which associations are affected by the change
+    if isinstance(obj, Contribution):
```
ThiefMaster, line 63: wanna use some [fancy modern python](https://peps.python.org/pep-0634/#class-patterns)? ;)

```py
match obj:
    case Contribution():
        ...
    case SessionBlock():
        ....
    case _:
        raise TypeError('Unsupported object')
```

zoom_rooms/indico_zoom_rooms/util.py
ThiefMaster: I just realized that `room.attributes` is either never falsy or triggers a query when used in a bool context. So not a great idea in either case.

But I think you can replace this *whole* construct with this:

```py
if room:
    return room.get_attribute_value('zoom-rooms-calendar-id')
```

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,130 @@
...
+            timeout=ZoomRoomsPlugin.settings.get('timeout'),
+        )
+    except requests.Timeout:
+        logger.warning('Request timed out')
+    except requests.RequestException:
+        logger.exception('%s request failed:\nURL: %s\nData: %s', method, url, pformat(data))
+    else:
+        logger.info('%s request to %s finished with status %r and body %r', method, path, res.status_code, res.text)
+        if res.ok:
+            return True
```
ThiefMaster, line 65: You could do `res.raise_for_status()` in the `try` block and then catch `requests.HTTPError` to handle non-successful status codes

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,130 @@
...
+from indico.core.db import db
+from indico.modules.events.contributions.models.contributions import Contribution
+from indico.modules.events.models.events import Event
+from indico.modules.events.sessions.models.blocks import SessionBlock
+from indico.util.string import strip_control_chars
+
+from indico_zoom_rooms.models import ZoomRoomsAction, ZoomRoomsQueueEntry
+
+
+def obj_from_locator(locator: dict) -> Contribution | SessionBlock | Event:
```
ThiefMaster, line 24: Isn't this a simplified version of `get_object_from_args` which we already have in the core? Assuming you have full locators that contain the parent objects.

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,270 @@
...
+        self.connect(signals.event.contribution_updated, _signal_link_object_updated)
+        self.connect(signals.event.session_block_updated, _signal_link_object_updated)
+        self.connect(signals.event.updated, _signal_link_object_updated)
+        self.connect(signals.event.times_changed, _signal_tt_entry_updated, sender=Contribution)
+        self.connect(signals.event.times_changed, _signal_tt_entry_updated, sender=SessionBlock)
+        self.connect(signals.vc.new_vc_room, _signal_zoom_meeting_created)
+        self.connect(signals.vc.cloned_vc_room, _signal_zoom_meeting_cloned)
+        self.connect(signals.vc.attached_vc_room, _signal_zoom_meeting_association_attached)
+        self.connect(signals.vc.detached_vc_room, _signal_zoom_meeting_association_detached)
+        self.connect(signals.vc.updated_vc_room_data, _signal_zoom_meeting_data_updated)
```
ThiefMaster, line 252: Usually we put those signal handlers as methods in the plugin class.

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,130 @@
...
+        return Contribution.get(locator['contrib_id'])
+    elif 'block_id' in locator:
+        return SessionBlock.get(locator['block_id'])
+    elif 'event_id' in locator:
+        return Event.get(locator['event_id'])
+    else:
+        raise ValueError("Can't decode locator")
+
+
+logger = get_task_logger(__name__)
```
ThiefMaster, line 36: I would rather use the Indico loggers for it, e.g. `current_plugin.logger` if this runs in the plugin context which I think it does. We want those log messages in `indico.log` and not `celery.log`

zoom_rooms/indico_zoom_rooms/plugin.py
pferreir: Yeah, I know. But is it really useful in this case?

ThiefMaster: Doesn't really make a difference I guess, so mostly consistency (and being able to use `self` isntead of `ZoomRoomsPlugin.instance`, but you're not using that in any of the signal handlers anyway).

zoom_rooms/indico_zoom_rooms/util.py
pferreir: Perfect. Thanks!

zoom_rooms/indico_zoom_rooms/util.py
pferreir: Maybe `chain.from_iterable(...)`. I'll try that.

zoom_rooms/indico_zoom_rooms/util.py
pferreir: Always ready to use fancy modern Python.

zoom_rooms/indico_zoom_rooms/tasks.py
pferreir: Seems like we're not using this function anymore. I'll get rid of it.

zoom_rooms/indico_zoom_rooms/tasks.py
pferreir: Gotcha. Done.

zoom_rooms/indico_zoom_rooms/plugin.py
pferreir: Maybe it's my inner Rustacean showing, but I tend to dislike methods which do not use `self` at all. And class methods in Python are very messy...

ThiefMaster: > And class methods in Python are very messy...

Agreed, I never liked certain linters (looking at you, PyCharm) suggesting people to make something a staticmethod just because it's not using self. IMHO something should be a classmethod/staticmethod when it's *meant to be called* in such a way, and not just because it's technically *possible* to call it w/o an instance...

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _send_request(method: str, zr_id: str, entry_id: str, data: dict | None = None) -> bool:
+    """Send a request to the corresponding HTTP service."""
+    from indico_zoom_rooms.plugin import ZoomRoomsPlugin
+    logger = current_plugin.logger
+
+    path = f'/api/v1/users/{zr_id}/events/{entry_id}'
+    if not (url := ZoomRoomsPlugin.settings.get('service_url')):
+        raise KeyError('service_url is not set!')
```
ThiefMaster, line 28: KeyError? Feels a bit weird here...

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+    url = url.rstrip('/') + path
```
ThiefMaster, line 29: Not tested, but maybe it's cleaner to use urllib for this? Assuming you never expect another path in the service URL of course.

```python
url = urlunsplit(urlsplit(url)._replace(path=path))
```

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+    return _send_request('DELETE', zr_id, entry_id)
+
+
+@celery.periodic_task(run_every=crontab(minute='*/1'), plugin='vc_zoom')
+def update_zoom_rooms_calendar_entries():
+    """Periodic task which sends all queued up entries and sends them to the HTTP API."""
+    to_delete = set()
+    logger = current_plugin.logger
+
+    for entry in ZoomRoomsQueueEntry.query.all():
```
ThiefMaster, line 87: Does this need any sorting?

zoom_rooms/indico_zoom_rooms/templates/explanation.html
```diff
@@ -0,0 +1,7 @@
+<a
+    href="{{ docs_url }}"
+    style="margin: 0;"
+    class="ui basic mini blue button"
+    title="{% trans %}Room &quot;{{room_name}}&quot; is equipped with a Zoom Room. Click to find out more.{% endtrans %}">
```
ThiefMaster, line 5:
```suggestion
    title="{% trans %}Room &quot;{{ room_name }}&quot; is equipped with a Zoom Room. Click to find out more.{% endtrans %}">
```

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,271 @@
...
+        [DataRequired()],
+        toggle=True,
+        description=_('The token used to authenticate with the CERN calendar service'),
+    )
+    timeout = FloatField(_('Request timeout'), [NumberRange(min=0.25)], description=_('Request timeout in seconds'))
+
+
+def only_zoom_rooms(f):
+    """Limit the wrapped handler to only objects which are somehow tied to a VCRoom and a Room with
+       ZoomRooms enabled."""
```
ThiefMaster, line 53:
```suggestion
    """
    Limit the wrapped handler to only objects which are somehow tied to a VCRoom and a Room with
   ZoomRooms enabled.
   """
```

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,271 @@
...
+            ZoomRoomsQueueEntry.record(ZoomRoomsAction.create, zr_id, assoc=assoc)
+
+
+@only_zoom_rooms
+def _handle_link_object_dt_change(
+    obj: SessionBlock | Contribution | Event,
+    changes: dict[str, t.Any],
+):
+    """Handle the changes of date/time of an object."""
+    if zr_id := get_zoom_room_id(obj.room):
```
ThiefMaster, line 90:
```py
if not (...):
    return
...
```

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,271 @@
...
+            # this doesn't concern us as there are no zoom rooms involved
+            return
+        case None, zr_id:
+            make_op = lambda assoc: ZoomRoomsQueueEntry.record(ZoomRoomsAction.create, t.cast(str, zr_id), assoc=assoc)
+        case zr_id, None:
+            make_op = lambda assoc: ZoomRoomsQueueEntry.record(ZoomRoomsAction.delete, t.cast(str, zr_id), assoc=assoc)
+        case old_zr_id, new_zr_id:
+            make_op = lambda assoc: ZoomRoomsQueueEntry.record(
+                ZoomRoomsAction.move, t.cast(str, old_zr_id), assoc=assoc, args={'new_zr_id': t.cast(str, new_zr_id)}
+            )
```
ThiefMaster, line 129:
```suggestion
        case old_id, new_id:
            make_op = lambda assoc: ZoomRoomsQueueEntry.record(
                ZoomRoomsAction.move, t.cast(str, old_id), assoc=assoc, args={'new_zr_id': t.cast(str, new_id)}
            )
```

I don't feel really strongly about it, but it took me a moment to notice htat it's just the catchall case where it re-assigns the same values to the same names...

zoom_rooms/indico_zoom_rooms/plugin.py
ThiefMaster: Hm, what about moving the functions into a separate files (handlers.py) like we did for livesync? Keeps the plugin file cleaner as well!

zoom_rooms/indico_zoom_rooms/plugin_test.py
```diff
@@ -0,0 +1,841 @@
...
+def test_move_between_blocks(mock_tasks, zr_session):
+    mock_tasks.reset()
+
+    (
+        _session,
+        block_1,
+        block_2,
+        contrib_1,
+        _contrib_2,
+    ) = zr_session
```
ThiefMaster, line 651: Much less ugly IMHO

```suggestion
    _session, block_1, block_2, contrib_1, _contrib_2 = zr_session
```

zoom_rooms/indico_zoom_rooms/plugin_test.py
```diff
@@ -0,0 +1,841 @@
...
+        contrib_2,
+    )
+
+
+EVENT_ENTRY_ID = 'zoom_meeting:test-zoom-room-1@example.com@indico:event:1#zmeeting1'
+ENTRY_TITLE = 'Zoom Meeting'
+
+
+def test_event(zr_event, mock_tasks, dummy_room, zr_powered_room):
+    tasks = mock_tasks()
```
ThiefMaster, line 196: Is this needed at the beginning of the test?

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,271 @@
...
+            return
+
+        if zr_id := get_zoom_room_id(event_vc_room.link_object):
+            return render_plugin_template(
+                'explanation.html',
+                room_name=event_vc_room.link_object.room.full_name,
+                event_vc_room=event_vc_room,
+                zr_id=zr_id,
+                docs_url=ZoomRoomsPlugin.settings.get('docs_url'),
+                **kwargs,
```
ThiefMaster, line 270: I would not pass along random kwargs - the signal handler needs to accept them to be forwards-compatible in case we start passing anything new, but since your template does not expect anything besides the `event_vc_room`, I don't think it's a good idea to pass extra ones.

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+        },
+    )
+
+
+def delete_entry(zr_id: str, entry_id: str) -> bool:
+    """Trigger a DELETE request."""
+    return _send_request('DELETE', zr_id, entry_id)
+
+
+@celery.periodic_task(run_every=crontab(minute='*/1'), plugin='vc_zoom')
```
ThiefMaster, line 81:
```suggestion
@celery.periodic_task(run_every=crontab(), plugin='vc_zoom')
```

Or if you want to be explicit, `minute='*'` - but all those values default to `*` anyway...

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,271 @@
...
+    category = PluginCategory.videoconference
+
+    def init(self):
+        super().init()
+        # Here we plug the various object signals into the action tracking code
+        # Consult the README.md for a summary of the logic behind
+        self.connect(signals.event.contribution_updated, _signal_link_object_updated)
+        self.connect(signals.event.session_block_updated, _signal_link_object_updated)
+        self.connect(signals.event.updated, _signal_link_object_updated)
+        self.connect(signals.event.times_changed, _signal_tt_entry_updated, sender=Contribution)
```
ThiefMaster, line 247: Does it need to be handled for `sender=Event` as well?

pferreir: No, because that's already caught by `signals.event.updated` (line above).

zoom_rooms/indico_zoom_rooms/tasks.py
pferreir: I'd rather leave it as `minute='*'`. Not everyone is aware that the minimum granularity is 1min and being explicit is better in this case.

zoom_rooms/indico_zoom_rooms/models.py
```diff
@@ -0,0 +1,137 @@
...
+            # DELETE operations have no additional parameters, only the ID
+            entry_data=None if action == ZoomRoomsAction.delete else EntryData(get_entry_data(obj, vc_room), **args),
+            extra_args=extra_args,
+            zoom_room_id=zoom_room_id,
+        )
+        db.session.add(entry)
+        db.session.flush()
+
+    @property
+    def data(self) -> dict:
```
ThiefMaster, line 130: Am I blind or is this used ONLY in the tests? If yes, I'd rather make this a `_entry_data` function in the in the file containing the tests, instead of masking it part of the actual model. It feels dirty to have code used only in tests...

zoom_rooms/indico_zoom_rooms/util.py
```diff
@@ -0,0 +1,88 @@
...
+    event_id = obj.id if isinstance(obj, Event) else obj.event.id
+    sub_id = ''
+    if isinstance(obj, Contribution):
+        sub_id = f':contribution:{obj.id}'
+    elif isinstance(obj, SessionBlock):
+        sub_id = f':block:{obj.id}'
+
+    zoom_id = vc_room.data['zoom_id']
+
+    return f'zoom_meeting:{user_id}@indico:event:{event_id}{sub_id}#{zoom_id}'
```
ThiefMaster, line 42: Is `#` really a good idea for something that becomes part of a URL path?

zoom_rooms/indico_zoom_rooms/plugin.py
ThiefMaster: I just tested and it's indeed the case. So it needs to be handled for `Event` as well, and possibly the `_signal_link_object_updated` handler should ignore time-related updates to avoid recording them twice in cases where both signals are triggered.

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,271 @@
...
+class PluginSettingsForm(IndicoForm):
+    debug = BooleanField(
+        _('Debug mode'),
+        widget=SwitchWidget(),
+        description=_('If enabled, requests are not sent to the API but logged instead'),
+    )
+    service_url = URLField(
+        _('Service URL'), [URL(require_tld=False)], description=_('The URL of the CERN calendar service')
+    )
+    docs_url = URLField(_('Documentation URL'), [URL()], description=_('The URL the "Zoom Rooms" buttons will link to'))
```
ThiefMaster, line 41: This should either be `DataRequired()` to show up as required (the `URL()` validator does not allow empty URLs, but it's weird to not have the field be shown as required), or `Optional()`; in the latter case the button should only be shown if there's a URL.

zoom_rooms/indico_zoom_rooms/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+from indico.core.plugins import current_plugin
+from indico.util.string import strip_control_chars
+
+from indico_zoom_rooms.models import ZoomRoomsAction, ZoomRoomsQueueEntry
+
+
+def _send_request(method: str, zr_id: str, entry_id: str, data: dict | None = None) -> bool:
+    """Send a request to the corresponding HTTP service."""
+    from indico_zoom_rooms.plugin import ZoomRoomsPlugin
+    logger = current_plugin.logger
```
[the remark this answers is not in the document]
ThiefMaster: in fact during testing i noticed that the `current_plugin` is the vc_zoom plugin, so yet another reason to use the `ZoomRoomsPlugin.logger`

zoom_rooms/indico_zoom_rooms/tasks.py
ThiefMaster: Actually I see we do the same in the outlook plugin, so let's keep it simple here as well

zoom_rooms/indico_zoom_rooms/models.py
pferreir: It is. I kept it because it might be used in the future. But I agree with you, let's move it to a less visible place.

zoom_rooms/indico_zoom_rooms/models.py
```diff
@@ -0,0 +1,145 @@
...
+    )
+
+    def __repr__(self):
+        action = ZoomRoomsAction(self.action).name
+        return format_repr(self, 'id', 'entry_id', action=action)
+
+    @classmethod
+    def record(
+        cls,
+        action: int,
```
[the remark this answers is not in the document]
pferreir: Indeed. Missed that one.

zoom_rooms/indico_zoom_rooms/models.py
```diff
@@ -0,0 +1,145 @@
...
+class ZoomRoomsQueueEntry(db.Model):
+    """Pending calendar updates"""
+
+    __tablename__ = 'queue'
+    __table_args__ = (
+        db.CheckConstraint(
+            f'action != {ZoomRoomsAction.delete} OR (entry_data IS NULL AND extra_args IS NULL)', 'delete_has_no_args'
+        ),
+        db.CheckConstraint(f'action = {ZoomRoomsAction.delete} OR entry_data IS NOT NULL', 'other_actions_have_args'),
+        db.CheckConstraint(f'action = {ZoomRoomsAction.move} OR extra_args IS NULL', 'move_has_extra_args'),
```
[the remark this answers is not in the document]
pferreir: Indeed. In the previous formulation I was only making sure that if the action was not a move, then there would be no arguments. Thanks.

zoom_rooms/indico_zoom_rooms/plugin.py
pferreir: Yes, you are right. I hadn't thought of that case. Ufff.

zoom_rooms/indico_zoom_rooms/models.py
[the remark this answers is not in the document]
pferreir: Ah, `ZoomRoomsAction` is not a true enum. It would have to inherit from `enum.Enum`.

zoom_rooms/indico_zoom_rooms/plugin_test.py
pferreir: hm... how do I get the tasks iterator otherwise?

zoom_rooms/indico_zoom_rooms/tasks.py
pferreir: Indeed. I believe it just works ATM because Postrgres returns everything in modification order by default (undocumented behaviour).

zoom_rooms/indico_zoom_rooms/tasks.py
pferreir: What would you recommend? I chose `KeyError` since it's a miss configuration option, could be seen as a missing key in a dict.

zoom_rooms/indico_zoom_rooms/plugin_test.py
ThiefMaster: huh i could've sworn I commented this on some reset call...

zoom_rooms/indico_zoom_rooms/plugin.py
pferreir: Not sure what you mean here. Is it just the name?

ThiefMaster: yeah, since you're basically doing the equivalent of `old_zr_id = old_zr_id` for that block. but like i said, no strong feeling about it, not sure what's actually cleaner

pferreir: Interesting, to me having the same name actually helps me realize it's a 1-to-1 mapping. Also no strong opinion. An alternative could be:

```python
       case _:
            make_op = lambda assoc: ZoomRoomsQueueEntry.record(
                ZoomRoomsAction.move, t.cast(str, old_zr_id), assoc=assoc, args={'new_zr_id': t.cast(str, new_zr_id)}
            )
```
But TBH it has its shortcomings too...

zoom_rooms/indico_zoom_rooms/util.py
pferreir: Good point. I can just replace it with another colon.

zoom_rooms/indico_zoom_rooms/plugin.py
pferreir: Great idea.

zoom_rooms/indico_zoom_rooms/plugin.py
pferreir: I think `times_changed` is only called in `track_time_changes`, so there's no need to do deduplication.

ThiefMaster: What I mean is that when you edit the event time via the general event settings, both signals get triggered. But removing times from the event updated signals also has the advantage that people are forced to use the correct signal.

zoom_rooms/indico_zoom_rooms/plugin.py
```diff
@@ -0,0 +1,88 @@
...
+        self.connect(signals.event.session_block_updated, handlers._signal_link_object_updated)
+        self.connect(signals.event.updated, handlers._signal_event_updated)
+        self.connect(signals.event.times_changed, handlers._signal_tt_entry_updated, sender=Contribution)
+        self.connect(signals.event.times_changed, handlers._signal_tt_entry_updated, sender=SessionBlock)
+        self.connect(signals.event.times_changed, handlers._signal_tt_entry_updated, sender=Event)
+        self.connect(signals.vc.vc_room_created, handlers._signal_zoom_meeting_created)
+        self.connect(signals.vc.vc_room_cloned, handlers._signal_zoom_meeting_cloned)
+        self.connect(signals.vc.vc_room_attached, handlers._signal_zoom_meeting_association_attached)
+        self.connect(signals.vc.vc_room_detached, handlers._signal_zoom_meeting_association_detached)
+        self.connect(signals.vc.vc_room_data_updated, handlers._signal_zoom_meeting_data_updated)
```
ThiefMaster, line 71: I'd ditch the leading underscores for those functions used from here (AFAIK vscode shows them as unused when they starts with an underscore and are not used in the same file)

PR #178, opened 2024-10-15
Draft: Unique res code

### LINKED ISSUE #177: CERNAccess: Enforce uniqueness of reservation codes
- Check if a generated code is already in the DB, and generate a new one if it is
- Add unique constraint/index to the DB
- Possibly replace the empty string with None when there is no code yet (that way you can use `unique=True` instead of requiring an unique index w/ a WHERE criterion to ignore empty strings)

AjobK:
Closes #177

### DIFF
--- a/cern_access/indico_cern_access/migrations/20241015_1702_92377810f14e_modify_reservation_code_field.py
+++ b/cern_access/indico_cern_access/migrations/20241015_1702_92377810f14e_modify_reservation_code_field.py
@@ -1,0 +1,29 @@
+"""Modify reservation code field
+
+Revision ID: 92377810f14e
+Revises: ee88b64f9494
+Create Date: 2024-10-15 17:02:59.886902
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+
+# revision identifiers, used by Alembic.
+revision = '92377810f14e'
+down_revision = 'ee88b64f9494'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_unique_constraint(
+        'uq_reservation_code', 'access_requests', ['reservation_code'],
+        schema='plugin_cern_access')
+
+
+def downgrade():
+    op.drop_constraint(
+        'uq_reservation_code', 'access_requests',
+        type_='unique', schema='plugin_cern_access')
--- a/cern_access/indico_cern_access/models/access_requests.py
+++ b/cern_access/indico_cern_access/models/access_requests.py
@@ -39,7 +39,8 @@ class CERNAccessRequest(db.Model):
     )
     reservation_code = db.Column(
         db.String,
-        nullable=False
+        nullable=False,
+        unique=True
     )
     adams_nonce = db.Column(
         db.String,
--- a/cern_access/indico_cern_access/util.py
+++ b/cern_access/indico_cern_access/util.py
@@ -269,7 +269,15 @@ def withdraw_event_access_request(req):
 def get_random_reservation_code():
     """Generate random reservation code for data required by ADaMS API."""
     charset = 'ABCDEFGHIJKLMNPQRSTUVWXYZ123456789'
-    return 'I' + ''.join(random.sample(charset, 6))
+    reservation_code = ''
+
+    while (
+        not reservation_code or
+        CERNAccessRequest.query.filter_by(reservation_code=reservation_code).first()
+    ):
+        reservation_code = 'I' + ''.join(random.sample(charset, 6))
+
+    return reservation_code
 
 
 def create_access_request(registration, state, reservation_code, nonce):


cern_access/indico_cern_access/migrations/20241015_1702_92377810f14e_modify_reservation_code_field.py
```diff
@@ -0,0 +1,29 @@
+"""Modify reservation code field
```
ThiefMaster, line 1: That does not describe what the revision does, better use something like "Make reservation code unique" (also rename the file accordingly).

cern_access/indico_cern_access/migrations/20241015_1702_92377810f14e_modify_reservation_code_field.py
```diff
@@ -0,0 +1,29 @@
...
+revision = '92377810f14e'
+down_revision = 'ee88b64f9494'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_unique_constraint(
+        'uq_reservation_code', 'access_requests', ['reservation_code'],
+        schema='plugin_cern_access')
```
ThiefMaster, line 23:
```suggestion
    op.create_unique_constraint(None, 'access_requests', ['reservation_code'], schema='plugin_cern_access')
```

the name is autogenerated by the sqlalchemy naming policy in the core

cern_access/indico_cern_access/migrations/20241015_1702_92377810f14e_modify_reservation_code_field.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def downgrade():
+    op.drop_constraint(
+        'uq_reservation_code', 'access_requests',
+        type_='unique', schema='plugin_cern_access')
```
ThiefMaster, line 29:
```suggestion
    op.drop_constraint('uq_reservation_code', 'access_requests', schema='plugin_cern_access')
```

- fits in one line (and the wrapping style was weird :D)
- we're using a decent DB so repeating the type here is not needed (iirc it's mysql or sqlite that requires it)

cern_access/indico_cern_access/util.py
```diff
@@ -269,7 +269,15 @@ def withdraw_event_access_request(req):
 def get_random_reservation_code():
     """Generate random reservation code for data required by ADaMS API."""
     charset = 'ABCDEFGHIJKLMNPQRSTUVWXYZ123456789'
-    return 'I' + ''.join(random.sample(charset, 6))
+    reservation_code = ''
+
+    while (
```
ThiefMaster, line 274:
```
while True:
    code = ...
    if <not unique>:
        break
```

I think that's nicer. It also most likely lets you avoid wrapping the if criterion

cern_access/indico_cern_access/util.py
```diff
@@ -269,7 +269,15 @@ def withdraw_event_access_request(req):
...
+        not reservation_code or
+        CERNAccessRequest.query.filter_by(reservation_code=reservation_code).first()
```
ThiefMaster, line 276: `.has_rows()` since you just care about existence, not about the actual data

cern_access/indico_cern_access/util.py
AjobK: Good suggestion, and actually instead of breaking I'll just return the reservation_code right away

cern_access/indico_cern_access/migrations/20241015_1702_92377810f14e_make_reservation_code_unique_and_nullable.py
```diff
@@ -0,0 +1,25 @@
...
+# revision identifiers, used by Alembic.
+revision = '92377810f14e'
+down_revision = 'ee88b64f9494'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_unique_constraint(None, 'access_requests', ['reservation_code'], schema='plugin_cern_access')
+    op.alter_column('access_requests', 'reservation_code', nullable=True, schema='plugin_cern_access')
```
ThiefMaster, line 20: also UPDATE existing rows and change from `''` to `NULL`

cern_access/indico_cern_access/migrations/20241015_1702_92377810f14e_make_reservation_code_unique_and_nullable.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def downgrade():
```
ThiefMaster, line 23: and vice versa here

cern_access/indico_cern_access/models/access_requests.py
```diff
@@ -39,7 +39,7 @@ class CERNAccessRequest(db.Model):
     )
     reservation_code = db.Column(
         db.String,
-        nullable=False
```
ThiefMaster, line 42: I'd rather be explicit and put `nullable=True` here instead of falling back to the implicit default

ThiefMaster (approved):

PR #180, opened 2024-10-24
Fix merchant reference to only contain printable ascii

tomasr8:
Postfinance requires that merchant reference only contains printable ascii characters.

### DIFF
--- a/payment_cern/indico_payment_cern/tests/util_test.py
+++ b/payment_cern/indico_payment_cern/tests/util_test.py
@@ -17,6 +17,10 @@ from indico_payment_cern.util import get_order_id
     (123, 456, 'x', 'Foo Bar',                         'XBARFOOc123r456'),
     (123, 456, '',  'FooVeryLongName BarVeryLongName', 'BARVERYLONGNAMEFOOVERYc123r456'),
     (123, 456, 'x', 'FooVeryLongName BarVeryLongName', 'XBARVERYLONGNAMEFOOVERc123r456'),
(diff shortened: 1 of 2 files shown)


payment_cern/indico_payment_cern/tests/util_test.py
```diff
@@ -17,6 +17,10 @@
     (123, 456, 'x', 'Foo Bar',                         'XBARFOOc123r456'),
     (123, 456, '',  'FooVeryLongName BarVeryLongName', 'BARVERYLONGNAMEFOOVERYc123r456'),
     (123, 456, 'x', 'FooVeryLongName BarVeryLongName', 'XBARVERYLONGNAMEFOOVERc123r456'),
+    (123, 456, 'x', 'Björn Groß', 'XGROSSBJOERNc123r456'),
+    (123, 456, 'x', 'Søren Åkesson', 'XAAKESSONSORENc123r456'),
+    (123, 456, 'x', 'Tomáš Dvořák', 'XDVORAKTOMASc123r456'),
+    (123, 456, 'x', 'けんじ 中村', 'Xc123r456'),  # XXX: This results in an empty order ID
```
tomasr8, line 23: Should we do something about this?

ThiefMaster: You still have the event/registration id ref so it's fine. Not much we can do beyond that anyway...

ThiefMaster (approved):

PR #182, opened 2024-11-14
Hide status overrides if plugin isn't active

alexiri:


ThiefMaster:
Added a core fix to make this work: indico/indico@e63bce8b88

PR #183, opened 2024-11-15
Allow custom reminders/times

alexiri:
Let the user configure if they want reminders or not, and how far in advance. Also let them define those as overrides per category.

@ThiefMaster note that this still needs some work, and I'm hoping you can give me a hint. My idea for the overrides is to have a single field where the user could define the status, reminder and/or reminder_time per category. However, I can't figure out how to get MultipleItemsField to have a validator so that _at least one of_ those values is filled in. Also, if the user doesn't specify a value for a particular variable, it should be passed as a None, otherwise  `_get_status`/`_get_reminder` can't tell the difference between an actual override and a blank. Perhaps this is too complicated for `MultipleItemsField` as it's currently implemented?

### DIFF
--- a/outlook/indico_outlook/plugin.py
+++ b/outlook/indico_outlook/plugin.py
@@ -80,7 +80,8 @@ class OutlookUserPreferences(ExtraUserPreferences):
                 {'id': 'id', 'caption': _('Category ID'), 'required': True, 'type': 'number', 'step': 1, 'coerce': int},
                 {'id': 'status', 'caption': _('Status'), 'required': False, 'type': 'select'},
                 {'id': 'reminder', 'caption': _('Reminder'), 'required': False, 'type': 'checkbox'},
-                {'id': 'reminder_minutes', 'caption': _('Reminder time'), 'required': False, 'type': 'number', 'step': 1, 'coerce': int},
+                {'id': 'reminder_minutes', 'caption': _('Reminder time'), 'required': False,
+                    'type': 'number', 'step': 1, 'coerce': int},
             ],
             choices={
                 'type': {'category': _('Category'), 'category_tree': _('Category & Subcategories')},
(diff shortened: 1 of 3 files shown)


outlook/indico_outlook/plugin.py
```diff
@@ -122,7 +138,9 @@ class OutlookPlugin(IndicoPlugin):
     default_user_settings = {
         'enabled': True,  # XXX: if the default value ever changes, adapt `get_participating_users`!
         'status': None,
-        'status_overrides': [],
```
ThiefMaster, line 125: Renaming the setting will require an alembic revision that updates it in the database (hopefully it's backwards compatible, otherwise it needs either some Postgres JSONB magic to convert the data - or just query, do it in python and update one by one - perfectly fine here as well)

You can run `indico db --plugin outlook revision -m 'Update user setting'` to create an empty revision file, and to execute it run `indico db --plugin outlook upgrade`

outlook/indico_outlook/plugin.py
```diff
@@ -66,35 +66,51 @@ class OutlookUserPreferences(ExtraUserPreferences):
...
-            _('Outlook entry status overrides'),
+        'outlook_reminder': BooleanField(_('Reminder'), [HiddenUnless('extra_outlook_active',
+                                                preserve_data=True)],
+                                    description=_('Enable calendar reminder')),
+        'outlook_reminder_minutes': IntegerField(_('Reminder time'), [HiddenUnless('extra_outlook_active',
+                                                preserve_data=True), NumberRange(min=0)],
+                                    description=_('Remind users X minutes before the event')),
+        'outlook_overrides': MultipleItemsField(
+            _('Outlook entry overrides'),
             [HiddenUnless('extra_outlook_active', preserve_data=True)],
```
ThiefMaster, line 77: Actually, you could add a custom validator here to check for useless entries and reject them

ThiefMaster:
> Perhaps this is too complicated for MultipleItemsField as it's currently implemented?

Yes, it's a rather simple field. More fancy things typically use a react-based widget nowadays. But that's probably overkill here... And I think leaving a field empty using the default would be fine. There's nothing where overriding with an empty string makes sense. And for reminder minutes, 0 would be an override, and empty would be using the default. That's already possible.

I don't think "must fill at least one" validation is necessary either - the override simply won't do anything if all the fields are empty... Just like an override with an invalid category id won't do anything. (But see my other comment, doing the validation would probably not be hard either.)

outlook/indico_outlook/plugin.py
alexiri: I tried that, but that only works when you click save and not as you edit entries, which is nicer. Is that possible?

ThiefMaster: nope

outlook/indico_outlook/plugin.py
alexiri: Done!

outlook/indico_outlook/plugin.py
alexiri: I've side-stepped this issue by making all fields mandatory.

ThiefMaster (approved):
very nice PR!

PR #184, opened 2024-11-15
Add favorites to Outlook calendar

alexiri:
Add favorite events and categories to Outlook calendar.

For favorited categories, only new/updated events are added to the calendar because Indico is missing the signals `signals.users.favorite_category_added` and `signals.users.favorite_category_removed`.

### DIFF
--- a/outlook/indico_outlook/plugin.py
+++ b/outlook/indico_outlook/plugin.py
@@ -194,6 +194,11 @@ class OutlookPlugin(IndicoPlugin):
         # Users that have marked the event as favorite too
         users_to_update |= event.favorite_of
 
+        # Now look for users that have marked as favorite that event's category (or any of its parents)
+        for category in event.category.chain_query.all():
+            for user in category.favorite_of:
+                users_to_update.add(user)
+
         for user in users_to_update:
             self.logger.info('Event data change: updating %s in %r', user, event)
             self._record_change(event, user, OutlookAction.update)
--- a/outlook/indico_outlook/plugin.py
+++ b/outlook/indico_outlook/plugin.py
@@ -137,6 +137,8 @@ class OutlookPlugin(IndicoPlugin):
         self.connect(signals.event.deleted, self.event_deleted)
         self.connect(signals.core.after_process, self._apply_changes)
         self.connect(signals.users.merged, self._merge_users)
+        self.connect(signals.users.favorite_event_added, self.favorite_event_added)
+        self.connect(signals.users.favorite_event_removed, self.favorite_event_removed)
 
     def _extend_indico_cli(self, sender, **kwargs):
         @cli_command()
@@ -147,6 +149,14 @@ class OutlookPlugin(IndicoPlugin):
 
     def extend_user_preferences(self, user, **kwargs):
         return OutlookUserPreferences
+
+    def favorite_event_added(self, user, event, **kwargs):
+        self._record_change(event, user, OutlookAction.add)
+        self.logger.info('Favorite event added: updating %s in %r', user, event)
+
+    def favorite_event_removed(self, user, event, **kwargs):
+        self._record_change(event, user, OutlookAction.remove)
+        self.logger.info('Favorite event removed: updating %s in %r', user, event)
 
     def event_registration_state_changed(self, registration, **kwargs):
         if not registration.user:
@@ -178,7 +188,13 @@ class OutlookPlugin(IndicoPlugin):
     def event_updated(self, event, changes, **kwargs):
         if not changes.keys() & {'title', 'description', 'location_data'}:
             return
-        for user in get_participating_users(event):
+
+        # Registered users need to be informed about changes
+        users_to_update = set(get_participating_users(event))
+        # Users that have marked the event as favorite too
+        users_to_update |= event.favorite_of
+
+        for user in users_to_update:
             self.logger.info('Event data change: updating %s in %r', user, event)
             self._record_change(event, user, OutlookAction.update)
 
@@ -198,6 +214,12 @@ class OutlookPlugin(IndicoPlugin):
             return
         if 'outlook_changes' not in g:
             g.outlook_changes = []
+
+        if action == OutlookAction.remove:
+            # Only remove an event if the user *really* shouldn't have it in their calendar
+            if user in get_participating_users(event) or user in event.favorite_of:
+                return
+
         g.outlook_changes.append((event, user, action))
 
     def _apply_changes(self, sender, **kwargs):
--- a/outlook/indico_outlook/plugin.py
+++ b/outlook/indico_outlook/plugin.py
@@ -62,6 +62,9 @@ class OutlookUserPreferences(ExtraUserPreferences):
         'outlook_active': BooleanField(_('Sync with Outlook'), widget=SwitchWidget(),
                                        description=_('Add Indico events in which I participate to my Outlook '
                                                      'calendar')),
+        'outlook_favorites': BooleanField(_('Sync favorites with Outlook'), [HiddenUnless('extra_outlook_active',
+                                    preserve_data=True)], widget=SwitchWidget(),
+                                    description=_('Add events/categories I mark as favorite to my Outlook calendar')),
         'outlook_status': SelectField(_('Outlook entry status'), [HiddenUnless('extra_outlook_active',
                                                                                preserve_data=True)],
                                       choices=_status_choices,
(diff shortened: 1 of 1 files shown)


outlook/indico_outlook/plugin.py
```diff
@@ -62,6 +62,9 @@ class OutlookUserPreferences(ExtraUserPreferences):
         'outlook_active': BooleanField(_('Sync with Outlook'), widget=SwitchWidget(),
                                        description=_('Add Indico events in which I participate to my Outlook '
                                                      'calendar')),
+        'outlook_favorites': BooleanField(_('Sync favorites with Outlook'), [HiddenUnless('extra_outlook_active',
+                                    preserve_data=True)], widget=SwitchWidget(),
+                                    description=_('Add events/categories I mark as favorite to my Outlook calendar')),
```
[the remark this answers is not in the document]
alexiri: You want an extra setting for the current behavior? So 4 switches in a row?

outlook/indico_outlook/plugin.py
```diff
@@ -176,28 +198,62 @@ def event_registration_form_deleted(self, registration_form, **kwargs):
...
     def event_updated(self, event, changes, **kwargs):
-        if not changes.keys() & {'title', 'description', 'location_data'}:
+        if not changes.keys() & {'title', 'description', 'location_data', 'start_dt', 'end_dt', 'duration'}:
             return
-        for user in get_participating_users(event):
+
+        # Registered users need to be informed about changes
+        users_to_update = set(get_participating_users(event))
+        # Users that have marked the event as favorite too
+        users_to_update |= event.favorite_of
```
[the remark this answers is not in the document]
alexiri: :man_facepalming:

alexiri:
Regarding the concern of somebody marking Home as favorite, this is actually already an issue for the calendar exporter feature: you can export an ical file for the entire Indico instance. It takes a little while on the CERN instance and produces an ical file with 6114 events (for me, users with more access will probably get more), all ending in the future.

With the code as it's implemented now, marking a category as favorite wouldn't affect that many events because it's no longer fully recursive. Even if it were, though, this may not really be an issue depending on how `indico outlook` is executed. I had assumed it would be a celery task, but it doesn't look like it is and I can't find how it's triggered nor how often. I think as long as multiple executions aren't allowed to overlap, the impact of a large quantity of updates should be fairly limited.

ThiefMaster:
> Even if it were, though, this may not really be an issue depending on how indico outlook is executed. I had assumed it would be a celery task, but it doesn't look like it is and I can't find how it's triggered nor how often. I think as long as multiple executions aren't allowed to overlap, the impact of a large quantity of updates should be fairly limited.

It is. See `outlook/indico_outlook/__init__.py` (once every 15 minuts, and yes by default we lock tasks so they do not overlap). And it's not so much about not wanting the task to take long (it's already fairly slow as it is though!), but rather avoiding to accidentally add 100s or 1000s events to someone's calendar.

> Regarding the concern of somebody marking Home as favorite, this is actually already an issue for the calendar exporter feature: you can export an ical file for the entire Indico instance. It takes a little while on the CERN instance and produces an ical file with 6114 events (for me, users with more access will probably get more), all ending in the future.

Actually your comment just reminded me of the perfect solution!  We have something called visibility. Add a `.filter(Event.is_visible_in(favorite_category_id))` to the query for events (unfortunately this does not accept multiple categories, BUT you could possibly use the CTE for visibility information directly - check the code if `is_visible_in`!). I think visibility check AND access check together may be pretty reasonable to stop people from doing stupid things and flooding their calendar.

alexiri:
After 4386d9a,  this PR now depends on https://github.com/indico/indico/pull/6618.

outlook/indico_outlook/plugin.py
```diff
@@ -148,8 +178,73 @@ def outlook():
...
+            self.logger.info('Favorite category added: user %s added event %r', user, event)
+
+        self.logger.info('Favorite category removed: updating %s in %r', user, category)
+
+    def category_updated(self, category, changes, **kwargs):
+        if not changes.keys() & {'visibility', 'protection_mode'}:
+            return
+
+        # If the visibility changes, we should probably update all events in the categories
+        # that may suddenly become visible/invisible to users. Not quite sure how to handle this yet.
```
ThiefMaster, line 244: No, let's not do that... it's not a common operation, and also it would introduce a huge amount of complexity. Same with protection settings changes IMHO!

outlook/indico_outlook/plugin.py
```diff
@@ -148,8 +178,73 @@ def outlook():
...
+
     def event_registration_state_changed(self, registration, **kwargs):
-        if not registration.user:
+        if not (registration.user and self._user_tracks_registered_events(registration.user)):
```
ThiefMaster, line 247: IMHO this is clearer when reading the code:

```suggestion
        if not registration.user or not self._user_tracks_registered_events(registration.user):
```

outlook/indico_outlook/plugin.py
```diff
@@ -203,34 +297,94 @@ def event_registration_form_deleted(self, registration_form, **kwargs):
...
+            self._record_change(event, user, OutlookAction.force_remove)
 
     def _record_change(self, event, user, action):
         if is_event_excluded(event):
             return
         if 'outlook_changes' not in g:
             g.outlook_changes = []
+
+        if action == OutlookAction.remove:
+            # Only remove an event if the user *really* shouldn't have it in their calendar
```
ThiefMaster, line 373: Do we really need the new `force_remove` action? You could just add a new `force` kwarg to `_record_change`, and skip the checks here if it's set to True. No need to record this persistently.

outlook/indico_outlook/plugin.py
```diff
@@ -203,34 +297,94 @@ def event_registration_form_deleted(self, registration_form, **kwargs):
...
+
+            if user in get_registered_users(event) and self._user_tracks_registered_events(user):
+                return
+            if user in event.favorite_of and self._user_tracks_favorite_events(user):
+                return
+            for category in reversed(event.category.chain_query.all()):
+                if user in category.favorite_of \
+                    and self._user_tracks_favorite_categories(user) \
+                    and event.can_access(user):
+                    return
```
ThiefMaster, line 383: We don't use `\` wrapping in Indico:

```suggestion
                if (
                    user in category.favorite_of
                    and self._user_tracks_favorite_categories(user)
                    and event.can_access(user)
                ):
                    return
```

outlook/indico_outlook/plugin.py
```diff
@@ -203,34 +297,94 @@ def event_registration_form_deleted(self, registration_form, **kwargs):
...
+        if action == OutlookAction.remove:
+            # Only remove an event if the user *really* shouldn't have it in their calendar
+
+            if user in get_registered_users(event) and self._user_tracks_registered_events(user):
+                return
+            if user in event.favorite_of and self._user_tracks_favorite_events(user):
+                return
+            for category in reversed(event.category.chain_query.all()):
+                if user in category.favorite_of \
+                    and self._user_tracks_favorite_categories(user) \
```
ThiefMaster, line 381: I'd do this check earlier since you do not need to iterate over categories at all when the user did not enable favorite categories.

outlook/indico_outlook/plugin.py
```diff
@@ -203,34 +297,94 @@ def event_registration_form_deleted(self, registration_form, **kwargs):
...
+                    and event.can_access(user):
+                    return
+                # Stop once we reach the visibility horizon of the event
+                if category is event.category.real_visibility_horizon:
```
ThiefMaster, line 385: I'm almost certain we can do some magic to use a single query to check whether the event is visible in one of the user's favorite categories... I may have another look later.

outlook/indico_outlook/plugin.py
```diff
@@ -151,7 +178,10 @@ class OutlookPlugin(IndicoPlugin):
         'max_event_duration': TimedeltaConverter
     }
     default_user_settings = {
-        'enabled': True,  # XXX: if the default value ever changes, adapt `get_participating_users`!
+        'enabled': True,  # XXX: if the default value ever changes, adapt `get_registered_users`!
+        'registered': True,
+        'favorite_events': True,
+        'favorite_categories': True,
```
ThiefMaster, line 184:
```suggestion
        'favorite_events': False,
        'favorite_categories': False,
```

For events I'm unsure (defaulting to `False` avoids surprises for users), but categories certainly need to be opt-in.

outlook/indico_outlook/plugin.py
ThiefMaster: Yeah I'm having a look. Tricky for sure, also since it's been aces since I touched those black magic CTEs for category trees. Might not be worth it in the end...

outlook/indico_outlook/plugin.py
alexiri: I'm concerned about the discoverability of this feature, if it's disabled by default nobody will find it and it won't be very useful.

PR #186, opened 2024-11-22
Enrich the data of created events

alexiri:


outlook/indico_outlook/calendar.py
```diff
@@ -106,17 +107,35 @@ def _update_calendar_entry(entry, settings):
...
+        cal_description.append(event.description)
+        if event.external_url:
+            cal_description.append(f'<p><a href="{event.external_url}">{event.external_url}</a></p>')
+
+        # Try to add the VC room URL to the description, if there is one
+        for vc_room in event.vc_room_associations:
+            try:
+                cal_description.append(f'<p><a href="{vc_room.data.url}">{vc_room.data.url}</a></p>')
+            except AttributeError:
+                pass
```
ThiefMaster, line 129: `data` is a dict, so this will always go in the `except` block

alexiri: Oops, I got carried away copying from the javascript part.

outlook/indico_outlook/calendar.py
```diff
@@ -106,17 +106,35 @@ def _update_calendar_entry(entry, settings):
...
+                    if event.venue_name and event.room_name
+                    else (event.venue_name or event.room_name))
+
+        cal_description = []
+        if event.person_links:
+            speakers = [f'{x.full_name} ({x.affiliation})' if x.affiliation else x.full_name
+                    for x in event.person_links]
+            cal_description.append(f'<p>Speakers: {', '.join(speakers)}</p>')
+        cal_description.append(event.description)
+        if event.external_url:
```
ThiefMaster, line 120: This is never falsy

outlook/indico_outlook/calendar.py
```diff
@@ -106,17 +106,35 @@ def _update_calendar_entry(entry, settings):
...
-        event_url = event.external_url
+
+        location = (f'{event.room_name} ({event.venue_name})'
+                    if event.venue_name and event.room_name
+                    else (event.venue_name or event.room_name))
+
+        cal_description = []
+        if event.person_links:
+            speakers = [f'{x.full_name} ({x.affiliation})' if x.affiliation else x.full_name
+                    for x in event.person_links]
```
ThiefMaster, line 117:
```suggestion
            speakers = [f'{x.full_name} ({x.affiliation})' if x.affiliation else x.full_name
                        for x in event.person_links]
```

outlook/indico_outlook/calendar.py
```diff
@@ -106,17 +106,35 @@ def _update_calendar_entry(entry, settings):
...
+        cal_description.append(event.description)
+        if event.external_url:
+            cal_description.append(f'<p><a href="{event.external_url}">{event.external_url}</a></p>')
+
+        # Try to add the VC room URL to the description, if there is one
+        for vc_room in event.vc_room_associations:
+            try:
+                cal_description.append(f'<p><a href="{vc_room.data['url']}">{vc_room.data['url']}</a></p>')
+            except KeyError:
+                pass
```
ThiefMaster, line 128: This is already done below when calling the `event.metadata_postprocess.send` signal (and instead of just trying if there is a URL, it checks if it's actually supposed to be included - for example in case of Zoom it checks if you should be able to see the passcode and if yes includes it in the URL)

ThiefMaster (approved):

Issue #191, opened 2024-12-13
Adapt Conversion plugin to full CloudConvert usage


ThiefMaster:
- [x] ~~Add new setting to~~ Convert all files using CloudConvert, not only public ones
- [x] Use generic filename instead of the user-provided one when sending files to CloudConvert
- [x] Add optional information notice (content configurable via a setting) that's shown in the attachment upload dialog to inform people about the conversion using CloudConvert (this could be a somewhat generic disclaimer setting)

And for later (not part of this issue, just mentioning it for the sake of completeness):

- Remove Doconverter integration
- Remove settings related to which conversion service should be used

PR #192, opened 2025-01-27
Convert all files with CloudConvert

GovernmentPlates:
Closes: #191

### DIFF
--- a/conversion/indico_conversion/conversion.py
+++ b/conversion/indico_conversion/conversion.py
@@ -207,13 +207,14 @@ def submit_attachment_cloudconvert(task, attachment):
     }
 
     try:
+        file_ext = os.path.splitext(attachment.file.filename)[1]
         job = client.Job.create(payload=job_definition)
         upload_task = job['tasks'][0]
         export_task = job['tasks'][-1]
         assert upload_task['operation'] == 'import/upload'
         assert export_task['operation'] == 'export/url'
         with attachment.file.open() as fd:
-            client.Task.upload(upload_task, attachment.file.filename, fd, attachment.file.content_type)
+            client.Task.upload(upload_task, f'attachment{file_ext}', fd, attachment.file.content_type)
         # add polling in case we miss a webhook
         export_task_id = export_task['id']
         cloudconvert_task_cache.set(export_task_id, 'pending')
--- a/conversion/indico_conversion/plugin.py
+++ b/conversion/indico_conversion/plugin.py
@@ -42,7 +42,7 @@ class SettingsForm(IndicoForm):
                                description=_('Temporarily disable submitting files. The tasks will be kept and once '
                                              'this setting is disabled the files will be submitted.'))
     use_cloudconvert = BooleanField(_('Use CloudConvert'), widget=SwitchWidget(),
-                                    description=_('Use Cloudconvert instead of Doconverter for public materials'))
+                                    description=_('Use Cloudconvert instead of Doconverter for PDF conversion.'))
     server_url = URLField(_('Doconverter server URL'), [DataRequired()],
                           description=_("The URL to the conversion server's uploadFile.py script."))
     cloudconvert_api_key = IndicoPasswordField(_('CloudConvert API key'),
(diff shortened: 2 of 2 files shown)


conversion/indico_conversion/plugin.py
```diff
@@ -105,10 +109,14 @@ def get_vars_js(self):
 
     def _add_file_form_fields(self, form_cls, **kwargs):
         exts = ', '.join(self.settings.get('valid_extensions'))
+        description = _('If enabled, your files will be converted to PDF if possible. '
+                        'The following file types can be converted: {exts}').format(exts=exts)
+        if self.settings.get('cloudconvert_conversion_notice'):
+            description = '{}\n\n{}'.format(self.settings.get('cloudconvert_conversion_notice'),
```
ThiefMaster, line 115: I don't think those `\n` will be visible, will they?

GovernmentPlates: Nope, it's not. Okay, I guess we can use markdown for this (as discussed).

conversion/indico_conversion/plugin.py
```diff
@@ -112,11 +114,11 @@ def _add_file_form_fields(self, form_cls, **kwargs):
         description = _('If enabled, your files will be converted to PDF if possible. '
                         'The following file types can be converted: {exts}').format(exts=exts)
         if self.settings.get('cloudconvert_conversion_notice'):
-            description = '{}\n\n{}'.format(self.settings.get('cloudconvert_conversion_notice'),
+            description = '{}<br>{}'.format(self.settings.get('cloudconvert_conversion_notice'),
                                           _('The following file types can be converted: {exts}').format(exts=exts))
         return 'convert_to_pdf', \
                BooleanField(_('Convert to PDF'), widget=SwitchWidget(),
-                            description=description,
+                            description=Markup(description),  # noqa: RUF035
```
GovernmentPlates, line 121: Not quite sure about this one (that `# noqa: RUF035`)...

tomasr8: ofc the first thing that comes up on google when searching for that rule is an issue from Adrian :laughing: https://github.com/astral-sh/ruff/issues/14523

ThiefMaster: I'd say here the warning from ruff is valid. With the current code, you allow any HTML, scripts included.

You probably want something like this here:

```python
Markup('{}<br><br>{}').format(Markup(render_markdown(...)), _(...))
```

That way the translated string does not allow any HTML, and the custom text can only contain the HTML that was sanitized at the end of `render_markdown()`.

conversion/indico_conversion/conversion.py
```diff
@@ -207,13 +207,14 @@ def submit_attachment_cloudconvert(task, attachment):
     }
 
     try:
+        file_ext = os.path.splitext(attachment.file.filename)[1]
```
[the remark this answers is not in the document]
GovernmentPlates: What about `pathlib.PurePath(...).suffix`?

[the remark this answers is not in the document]
ThiefMaster: does it really make sense to use pathlib when dealing with just a filename?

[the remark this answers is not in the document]
GovernmentPlates: True - I was sticking with `os.path.splitext(...)` since we are already using it elsewhere.

conversion/indico_conversion/conversion.py
```diff
@@ -207,13 +207,14 @@ def submit_attachment_cloudconvert(task, attachment):
...
         job = client.Job.create(payload=job_definition)
         upload_task = job['tasks'][0]
         export_task = job['tasks'][-1]
         assert upload_task['operation'] == 'import/upload'
         assert export_task['operation'] == 'export/url'
         with attachment.file.open() as fd:
-            client.Task.upload(upload_task, attachment.file.filename, fd, attachment.file.content_type)
+            client.Task.upload(upload_task, f'attachment{file_ext}', fd, attachment.file.content_type)
```
tomasr8, line 217: Do we want to include the attachment id in the name?

GovernmentPlates: Nope - the idea is to not send any identifying data to the conversion service.

conversion/indico_conversion/plugin.py
```diff
@@ -105,10 +113,17 @@ def get_vars_js(self):
 
     def _add_file_form_fields(self, form_cls, **kwargs):
         exts = ', '.join(self.settings.get('valid_extensions'))
+        description = _('If enabled, your files will be converted to PDF if possible. '
```
ThiefMaster, line 116: You're currently losing the original description as soon as a notice is set

ThiefMaster:
The `use_cloudconvert` setting still mentions "for public materials"; that text should go away

PR #193, opened 2025-02-10
Ravem: Show correct name of connected VC

ThiefMaster:
Kind of ugly because the state information is lost when you click disconnected and then don't force the disconnect, but I really did not want to touch more of this somewhat legacy JS code than necessary.

### DIFF
--- a/ravem/indico_ravem/client/index.js
+++ b/ravem/indico_ravem/client/index.js
@@ -5,13 +5,16 @@
 // them and/or modify them under the terms of the MIT License; see
 // the LICENSE file for more details.
 
+/* eslint-disable func-name-matching */
+/* eslint-disable import/unambiguous */
+
 (function() {
   const $t = $T.domain('ravem');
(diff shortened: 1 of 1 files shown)


ravem/indico_ravem/client/index.js
```diff
@@ -192,7 +195,7 @@
     /**
      *  Sets a new state for the button and update its icon, label and tool tip.
      */
-    function setButtonState(btn, newState, tooltipMessage) {
+    function setButtonState(btn, newState, tooltipMessage = '', currentVCRoom = '') {
```
tomasr8, line 198: Does `tooltipMessage` need a default or can it be without?

ThiefMaster: it could be without, just feels nicer since we were not passing one before so it was just undefined anyway

tomasr8: yeah I don't have a strong opinion on it..

PR #199, opened 2025-07-29
Outlook: Store calendar entry UIDs in DB

ThiefMaster:
This will make dealing with favorite categories being synced in #184 much easier, because when removing a category from favorites we can simply query all events that were actually added to the user's calendar, instead of hammering the API with updates for every (future) event that *could* be in the user's calendar.

### DIFF
--- a/outlook/indico_outlook/calendar.py
+++ b/outlook/indico_outlook/calendar.py
@@ -18,6 +18,7 @@ from indico.core.db import db
 from indico.util.signals import values_from_signal
 from indico.util.string import strip_control_chars
 
+from indico_outlook.models.entry import OutlookCalendarEntry
 from indico_outlook.models.queue import OutlookAction, OutlookQueueEntry
 from indico_outlook.util import check_config, is_event_excluded
 
@@ -63,7 +64,7 @@ def update_calendar():
     finally:
         if delete_ids:
             OutlookQueueEntry.query.filter(OutlookQueueEntry.id.in_(delete_ids)).delete(synchronize_session=False)
-            db.session.commit()
+        db.session.commit()
 
 
 def _get_status(user, event, settings):
@@ -96,7 +97,13 @@ def _get_reminder(user, event, settings):
     return reminder, reminder_minutes
 
 
-def _update_calendar_entry(entry, settings):
+def _make_calendar_id(event, user, settings):
+    if settings['event_id_cutoff'] != -1 and event.id > settings['event_id_cutoff']:
+        return event.ical_uid
+    return f'{settings["id_prefix"]}{user.id}_{event.id}'
+
+
+def _update_calendar_entry(entry: OutlookQueueEntry, settings):
     """Executes a single calendar update
 
     :param entry: a :class:`OutlookQueueEntry`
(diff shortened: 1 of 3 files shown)


outlook/indico_outlook/calendar.py
```diff
@@ -96,7 +97,14 @@ def _get_reminder(user, event, settings):
     return reminder, reminder_minutes
 
 
-def _update_calendar_entry(entry, settings):
+def _make_calendar_id(event, user, settings):
+    if settings['event_id_cutoff'] != -1 and event.id > settings['event_id_cutoff']:
+        return event.ical_uid
+    else:
+        return f'{settings["id_prefix"]}{user.id}_{event.id}'
```
tomasr8, line 104: Maybe we can enable [RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/#superfluous-else-return-ret505)?

outlook/indico_outlook/calendar.py
```diff
@@ -183,6 +196,18 @@ def _update_calendar_entry(entry, settings):
         return False
     else:
         logger.info('Request to %s %s finished with status %r and body %r', method, path, res.status_code, res.text)
+        if res.ok and (entry.action == OutlookAction.add or (not existing and entry.action == OutlookAction.update)):
+            # successfully added or updated w/ no reference to existing entry
+            OutlookCalendarEntry.create(event, user, unique_id)
```
tomasr8, line 201: Is it worth creating a separate method for this considering it's just called once?

outlook/indico_outlook/calendar.py
ThiefMaster: I tend to prefer having both returns on the same level if they are "equally good".

If one branch errors out / aborts early I'd also return early and then outdent the rest, but for cases like this one I don't think there's a real benefit in outdenting the second return

ThiefMaster:
Just to have it archived here as well, this is the script to populate the table w/ recent data of existing calendar entries. We cannot easily get the data from the mail team (since their API just proxies requests to Exchange), but we have anything created/updated/deleted in the last 13 months, and the code is still resilient to missing entries (updates will just create the missing entry, and deletes will be ignored).

```sh
zgrep -h indico.plugin.outlook log/indico.log log/old/indico.log-*.gz | grep 'Request to' | sort | grep -oP "(?<=Request to ).+$" >! /tmp/outlook-requests.log
```

```py
import re
import sys

from indico.core.db import db
from indico.modules.events.models.events import Event
from indico.modules.users.models.users import User
from indico.util.console import verbose_iterator
from indico.web.flask.app import make_app

from indico_outlook.models.entry import OutlookCalendarEntry


LOG_LINE_RE = re.compile(
    r'^(?P<verb>[A-Z]+) /api/v1/users/(?P<email>[^/]+)/events/(?P<uid>[^ ]+) finished with status '
    r'(?P<code>\d+) and body \'(?P<body>.+)\'$',
    re.MULTILINE,
)
UID_RE = re.compile(r'^indico-event-(\d+)@indico\.cern\.ch$')
LEGACY_UID_RE = re.compile(r'^indico_\d+_(\d+)$')


def main():
    entries = set()
    print('Parsing log lines')
    for line in sys.stdin:
        if not (match := LOG_LINE_RE.match(line)):
            print(f'INVALID: {line}')
            break
        verb = match.group('verb')
        email = match.group('email')
        uid = match.group('uid')
        code = int(match.group('code'))
        assert verb in {'PUT', 'DELETE'}

        if 500 <= code < 600:
            # skip errors, those are retried anyway
            continue
        elif code == 400:
            # temporary bug, they were also retried
            continue

        key = (email, uid)
        match verb, code:
            case 'PUT', 200 | 201:  # entry created or updated
                entries.add(key)
            case 'PUT', 404:  # attempted create/update for invalid user
                entries.discard(key)
            case 'DELETE', 200 | 404:  # successful delete or invalid user
                entries.discard(key)
            case _:
                raise Exception(f'unexpected combination: {verb} {code}')

    entries = {(email, uid, _get_event_id(uid)) for email, uid in entries}
    emails = {x[0] for x in entries}
    event_ids = {x[2] for x in entries}
    print(f'Caching {len(event_ids)} events')
    event_cache = {e.id: e for e in Event.query.filter(Event.id.in_(event_ids))}
    print(f'Caching {len(emails)} users')
    user_cache = {}
    for user in User.query.filter(User.all_emails.in_(emails)).all():
        for email in user.all_emails:
            user_cache[email] = user

    seen = {}
    for email, uid, event_id in verbose_iterator(entries, len(entries)):
        if not (user := user_cache.get(email)):
            print(f'invalid user: {email}')
            continue

        if not (event := event_cache.get(event_id)):
            print(f'invalid event: {event_id}')
            continue

        key = (event, user)
        if key in seen:
            print(f'duplicate record for {event_id}: {email}, {seen[key]}')
            continue

        seen[key] = email
        OutlookCalendarEntry.create(event, user, uid)

    input('Press ENTER to commit')
    db.session.commit()


def _get_event_id(uid):
    if match := UID_RE.match(uid):
        return int(match.group(1))
    if match := LEGACY_UID_RE.match(uid):
        return int(match.group(1))
    raise Exception(f'unexpected uid: {uid}')


if __name__ == '__main__':
    with make_app().app_context():
        main()
```

